PR #64, opened 2024-07-01
Add CharacteristicBuilder and ServiceHandle

Univa:
The main motivation for this was to allow for adding descriptors to characteristics.

- Adds `CharacteristicBuilder`, which contains methods for adding descriptors
- Adds `ServiceHandle`, which can be obtained by calling a newly added `build` method to `ServiceBuilder`
- `ServiceBuilder::add_characteristic` and `ServiceBuilder::add_characteristic_ro` were changed to return a `CharacteristicBuilder` instead of `CharacteristicHandle`. `CharacteristicBuilder::build` can be called to obtain the handle.


lulf (approved):
Looks good, thanks!

PR #89, opened 2024-08-22
minor fixes to reduce unnecessary work

lulf:


host/src/attribute.rs
```diff
@@ -16,6 +16,8 @@ pub const CHARACTERISTIC_APPEARANCE_UUID16: Uuid = Uuid::Uuid16(0x2A03u16.to_le_
 pub const GENERIC_ATTRIBUTE_SERVICE_UUID16: Uuid = Uuid::Uuid16(0x1801u16.to_le_bytes());
 
 pub const PRIMARY_SERVICE_UUID16: Uuid = Uuid::Uuid16(0x2800u16.to_le_bytes());
+pub const SECONDARY_SERVICE_UUID16: Uuid = Uuid::Uuid16(0x2801u16.to_le_bytes());
+pub const INCLUDE_SERVICE_UUID16: Uuid = Uuid::Uuid16(0x2801u16.to_le_bytes());
```
8cichy8, line 20: Hello. I absolutely don't understand anything about BLE, but isn't there TYPO and shouldn't this be 0x2802 instead?

lulf: That's correct, thanks for the review! This got committed by mistake, and isn't actually in use.

lulf: Pushed a fix now

PR #97, opened 2024-08-26
Fix 128-bit UUID parsing in discover_characteristic

mertzt89:
This change addresses an issue where discovered characteristic UUID's were always being treated as 16-bit.


lulf (approved):
Thanks!

Issue #98, opened 2024-08-30
Add Attribute Data On Callback functionality


nm17:
## Current problem
Current GATT API requires the user to create a buffer for data that can't be changed afterwards by the outside code easily and requires the buffer to live as much as the device does (`'d`).

This limits the possible returned values from GATT to a fixed length, without modifying the Attribute Table, which isn't good developer experience.

## My proposed solution
Add two variants for AttributeData which allow read and write functionality, containing the props and the callback, which processes either reads or writes. I'm not exactly sure about the parameters and return type for the callback, I will think about that later. Any suggestions are welcome.

## The questions for now
  - Can I assign myself on something like this? I would like to attempt this myself.
  - Will this conflict with any other features which are currently in the works?
  - Is this even something that the devs of trouble would like to see implemented?

nm17:
Should this variant be done by:

  + Adding a new variant in AttributeData?  
     This option is easier to do, but will add another variant, which might be non ideal, since this enum is public and doesn't have `#[non_exhaustive]`, which will break API for those who match against it (?)
  + Somehow modifying the existing `Data` and `DataReadOnly` variants?  
     This one is harder, but will leave the amount of variants the same, without breaking compatibility. 

Honestly, either way is fine IMHO, since I don't think many people are matching on AttributeData either way.

lulf:
Thanks for raising the issue and sharing your ideas. I think storing the callback per attribute is not going to work generically without an allocator? But what could work is providing a callback to the gatt server perhaps, and then pass the handle and 'buffer' to the callback. 

Regarding breaking APIs I wouldn't worry about it at this point, we don't have any release on crates.io yet.

nm17:
I guess anything that allocates will be off limits for this repo? Or can I add a feature flag for this and put the additional AttributeData variant and functions that use it under it?

lulf:
Yes, I would like to avoid alloc. Especially if we can get almost similar functionality with just a slight inconvenience in the api.

lulf:
Maybe you could look at the gatt api for nrf softdevice for inspiration. It allows using heapless Vec types as attribute types, which allows for not always copying all the data.

lulf:
I'll throw another idea into the mix :) 

[...]

Issue #95, opened 2024-08-26
How to add descriptor to characteristic?


HaoboGu:
In current implementation, `add_descriptor` returns a `DescriptorHandle`, which doesn't have `build()` like `CharacteristicBuilder`. That means I cannot do things like:

```rust
let input_keyboard_handle = table
    .add_service(Service::new(BleSpecification::HidService as u16))
    .add_characteristic(
        BleCharacteristics::HidReport as u16,
        &[
            CharacteristicProp::Read,
            CharacteristicProp::Write,
            CharacteristicProp::Notify,
        ],
        &mut input_keyboard_data,
    )
    .add_descriptor(
        BleDescriptor::ReportReference as u16,
        &[
            CharacteristicProp::Read,
            CharacteristicProp::Write,
            CharacteristicProp::Notify,
        ],
        &mut input_keyboard_desc_data,
    )
    .build();
```

I can only get a `DescriptorHandle` without using `build`, which cannot be used to be compared using

```rust
  match server.next().await {
      Ok(GattEvent::Read {
          handle,
          connection: _,
      }) => {
          if handle == input_keyboard_handle {       <= ERROR: mismatched types: expected `Characteristic`, found `DescriptorHandle`
              info!("Input keyboard read");
          }
          ...
   }
```

lulf:
Thanks for raising the issue. The descriptor api is abit weird as it is, I'm not sure if it's really needed. I think you can work around it like this:

```
let mut svc = table
    .add_service(Service::new(BleSpecification::HidService as u16));

let input_kbd_data_h = svc.add_characteristic(
        BleCharacteristics::HidReport as u16,
        &[
            CharacteristicProp::Read,
            CharacteristicProp::Write,
            CharacteristicProp::Notify,
        ],
        &mut input_keyboard_data,
    ).build();

let input_kbd_desc_h = svc.add_descriptor(
        BleDescriptor::ReportReference as u16,
        &[
            CharacteristicProp::Read,
            CharacteristicProp::Write,
            CharacteristicProp::Notify,
        ],
        &mut input_keyboard_desc_data,
    )
    .build();
```

lulf:
I removed the api to add descriptor, you should just use the add_characteristic and add_characteristic_ro to add those to the ServiceBuilder instead.

HaoboGu:
thanks! I'll try it soon:D

HaoboGu:
I tried the latest commit, the following is my initialization code:

```rust
[...]

Issue #35, opened 2024-05-09
Documentation


lulf:
Document all public APIs and create a guide

lulf:
Documentation now published here https://embassy.dev/trouble/

PR #99, opened 2024-08-31
Work on the Attribute module

nm17:
Currently a draft, any comments are welcome!


host/src/attribute.rs
```diff
@@ -24,45 +24,85 @@ pub const GENERIC_ATTRIBUTE_UUID16: Uuid = Uuid::Uuid16(0x1801u16.to_le_bytes())
 
 #[derive(Debug, Clone, Copy)]
 #[repr(u8)]
+/// An enum of possible characteristic properties
+/// 
+/// Ref: BLUETOOTH CORE SPECIFICATION Version 6.0, Vol 3, Part G, Section 3.3.1.1 Characteristic Properties
```
nm17, line 29: I would like to propose this syntax for referencing Bluetooth core specification. This should help other devs cross reference things.

nm17:
Oh, and I forgot to `cargo fmt`. I will do it after, I just didn't want to screw something up.

nm17:
Next up I will try to document GattServer and stuff related to it. Then I will:

  - try to split the `AttributeData` read and write functionality into a separate trait
  - try to make `ServiceBuilder::add_characteristic_internal` public, but use the trait I added as a dyn type.

nm17:
Also, related issues: #98 #35

host/src/attribute.rs
```diff
@@ -204,8 +258,8 @@ impl<'d> AttributeData<'d> {
                     return Err(AttErrorCode::UnlikelyError);
                 }
 
-                *notifications = data[0] & 0x01 != 0;
-                *indications = data[0] & 0x02 != 0;
+                *notifications = data[0] & 0x00000001 != 0;
+                *indications = data[0] & 0x00000010 != 0;
```
igiona, line 262: This value is not quite the same as 0x02, was this intentional?

nm17: Oh yeah, I meant to do `0b`, not `0x`, my bad :sweat_smile:

nm17:
I changed a lot of stuff, so if anyone has the time, please review! :heart:

nm17:
I have no idea how the `error[E0597]: `bat_level` does not live long enough` error came to be in `example/apps/ble_bas_peripheral.rs`. I would like some help with this. I don't think that I've yet changed anything in how the `AttributeData` worked...

HaoboGu:
> I have no idea how the `error[E0597]: bat_level does not live long enough` error came to be in `example/apps/ble_bas_peripheral.rs`. I would like some help with this. I don't think that I've yet changed anything in how the `AttributeData` works...

I encountered similar error before, see #95 . But I'm not sure whether this is your case. There're some complex lifetime calculations.

.gitignore
```diff
@@ -0,0 +1,45 @@
...
+
+### Rust ###
+# Generated by Cargo
+# will have compiled files and executables
+debug/
+target/
+
+# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
+# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
+Cargo.lock
```
lulf, line 12: I'd like to not ignore the Cargo.lock for examples, so maybe you can create a specific .gitignore for the host/ folder?

host/src/attribute/mod.rs
```diff
@@ -1,216 +1,68 @@
...
-pub const CHARACTERISTIC_UUID16: Uuid = Uuid::Uuid16(0x2803u16.to_le_bytes());
-pub const CHARACTERISTIC_CCCD_UUID16: Uuid = Uuid::Uuid16(0x2902u16.to_le_bytes());
-pub const GENERIC_ATTRIBUTE_UUID16: Uuid = Uuid::Uuid16(0x1801u16.to_le_bytes());
-
-#[derive(Debug, Clone, Copy)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
 #[repr(u8)]
+/// An enum of possible characteristic properties
+///
+/// Ref: BLUETOOTH CORE SPECIFICATION Version 6.0, Vol 3, Part G, Section 3.3.1.1 Characteristic Properties
```
lulf, line 23: If we're going to refer to the speak, please use something like this, which makes it a lot easier to lookup and check:

https://github.com/embassy-rs/bt-hci/blob/main/src/event/le.rs#L182

host/src/attribute/mod.rs
```diff
@@ -1,216 +1,68 @@
...
+    /// Permit writes to the Characteristic Value without response
     Extended = 0x80,
 }
 
-pub struct Attribute<'a> {
+#[derive(PartialEq, Eq)]
+pub struct Attribute<'d> {
+    /// Attribute type UUID
+    ///
+    /// Do not mistake it with Characteristic UUID
```
lulf, line 50: I'm not sure if this is entirely correct, every characteristic has a UUID, but there are different types, like the characteristic declaration or the characteristic value. This UUID can be either.

host/src/attribute/mod.rs
```diff
@@ -1,216 +1,68 @@
...
     pub uuid: Uuid,
+    /// Handle for the Attribute
+    ///
+    /// In case of a push, this value is ignored and set to the
```
lulf, line 54: Change this to 'When constructing the attribute table'

host/src/attribute/mod.rs
```diff
@@ -1,216 +1,68 @@
...
-        handle: u16,
-        uuid: Uuid,
-    },
-    Cccd {
-        notifications: bool,
-        indications: bool,
-    },
+    /// Last handle value in the group
+    ///
+    /// When a [`ServiceBuilder`] finishes building, it returns the handle for the service, but also
```
lulf, line 59: Sentence ends prematurely.

host/src/types/uuid.rs
```diff
@@ -1,7 +1,7 @@
 use crate::codec::{Decode, Encode, Error, Type};
 
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[derive(Debug, PartialEq, Clone)]
+#[derive(Debug, PartialEq, Eq, Hash, Clone, Copy)]
```
lulf, line 4: I'd rather keep the uuid not Copy if possible, as it can easily explode memory usage if copied uncritically throughout the usage.

lulf (changes requested):
Thanks for the PR! I've had a first look through it, but the general set of changes look good. 

As for the lifetime issues, I think this is why I didn't use Vec originally, as it's lifetime would be shorter than the data that gets pushed to it which the compiler wouldn't accept. (Might be a way around it, maybe using some MaybeUninit trickery..)

PR #102, opened 2024-09-01
Add possibility to listen to notifications

pferreir:
This allows the `GattClient` to also register to listen to notifications for a Characteristic.

### DIFF
--- a/host/src/gatt.rs
+++ b/host/src/gatt.rs
@@ -1,16 +1,24 @@
+use core::cell::RefCell;
+use core::future::Future;
+use core::marker::PhantomData;
+
 use bt_hci::controller::Controller;
 use bt_hci::param::ConnHandle;
-use embassy_sync::blocking_mutex::raw::RawMutex;
-use embassy_sync::channel::{DynamicReceiver, DynamicSender};
+use embassy_sync::blocking_mutex::raw::{NoopRawMutex, RawMutex};
+use embassy_sync::channel::{Channel, DynamicReceiver, DynamicSender};
 use heapless::Vec;
 
 use crate::att::{self, AttReq, AttRsp, ATT_HANDLE_VALUE_NTF};
-use crate::attribute::{Characteristic, Uuid, CHARACTERISTIC_UUID16, PRIMARY_SERVICE_UUID16};
+use crate::attribute::{
+    AttributeData, Characteristic, CharacteristicProp, Uuid, CCCD, CHARACTERISTIC_CCCD_UUID16, CHARACTERISTIC_UUID16,
+    PRIMARY_SERVICE_UUID16,
+};
 use crate::attribute_server::AttributeServer;
 use crate::connection::Connection;
 use crate::connection_manager::DynamicConnectionManager;
 use crate::cursor::{ReadCursor, WriteCursor};
 use crate::host::BleHost;
+use crate::packet_pool::{GlobalPacketPool, Packet, ATT_ID};
 use crate::pdu::Pdu;
 use crate::types::l2cap::L2capHeader;
 use crate::{BleHostError, Error};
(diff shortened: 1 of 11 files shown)


host/src/gatt.rs
```diff
@@ -138,11 +142,35 @@ pub enum GattEvent<'reference> {
...
+
+pub struct GattClient<
+    'reference,
+    'resources,
+    T: Controller,
+    const MAX_SERVICES: usize,
+    const MAX_NOTIF: usize,
+    const NOTIF_QSIZE: usize,
+    const L2CAP_MTU: usize = 27,
+    const ATT_MTU: usize = 23,
```
lulf, line 166: These are a bit hard to keep in sync. I'd rather just use l2cap_mtu for the vec size to avoid them.

lulf (changes requested):
Thanks for the PR. Looks good in general.

Couple of things:

* I think maybe it's gonna explode RAM usage storing that Vec<> in the channels. Consider using the PacketPool instead, which should reduce the RAM requirements considerably, because at the moment you risk having a lot of 'unused RAM' from defining the MAX limits.
* I would also like if you modified the ble_bas_central example to use this.

pferreir:
> * I think maybe it's gonna explode RAM usage storing that Vec<> in the channels. Consider using the PacketPool instead, which should reduce the RAM requirements considerably, because at the moment you risk having a lot of 'unused RAM' from defining the MAX limits.

I think the problem with this is that PacketPool needs to have a static lifetime, and that means I will either have to ask the user to initialize that or require `'static` in some of the methods...

lulf:
> > * I think maybe it's gonna explode RAM usage storing that Vec<> in the channels. Consider using the PacketPool instead, which should reduce the RAM requirements considerably, because at the moment you risk having a lot of 'unused RAM' from defining the MAX limits.
> 
> I think the problem with this is that calls to `PacketPool::alloc` need to have a static lifetime, and that means I'll have to require `&'static self` in some of the methods...


You'd need to use a StaticCell for the pool, and pass it to the gatt_client when creating, it doesn't mean the gatt client itself must be static, only the Packet, i.e. Packet<'static>. Then it can assume static lifetime of an allocated packet. This is how the l2cap channel manager works (https://github.com/embassy-rs/trouble/blob/main/host/src/channel_manager.rs#L34-L41), where it holds a ref to a 'static pool, but doesn't require itself to be 'static.

There may very well be other pool implementations that is a better fit, but this one is at least already there with no new dependencies needed.

pferreir:
> You'd need to use a StaticCell for the pool, and pass it to the gatt_client when creating

Yeah, I had excluded this because it seemed not very user-friendly. But I've done it and it seems to work.

pferreir:
OK, I think this is more or less what you asked for. I made a design decision which I am not 100% sure of, which is to make the task which handles rx mandatory, so that messages can be triaged and notifications be handled in parallel with regular read/write operations. That also entailed making the service list internally mutable. I created a `ble_bas_central` example for the rp2040 to demonstrate the concept.
Let me know what you think!

pferreir:
BTW, haven't found anything to test the official example on. I have a device which has a battery service but I don't think it sends any notifications. I know the notification code works because I tested it with a BLE MIDI device I have, but I don't know what to do with the example.

host/src/gatt.rs
```diff
@@ -138,11 +146,49 @@ pub enum GattEvent<'reference> {
...
+    const NOTIF_QSIZE: usize,
+    const L2CAP_MTU: usize = 27,
+> {
+    pub(crate) known_services: RefCell<Vec<ServiceHandle, MAX_SERVICES>>,
     pub(crate) rx: DynamicReceiver<'reference, (ConnHandle, Pdu)>,
     pub(crate) ble: &'reference BleHost<'resources, T>,
     pub(crate) connection: Connection<'reference>,
+    pub(crate) request_channel: Channel<NoopRawMutex, (ConnHandle, Pdu), NOTIF_QSIZE>,
+
+    pub(crate) notification_pool: &'static PacketPool<NoopRawMutex, L2CAP_MTU, MAX_NOTIF, 1>,
```
lulf, line 189: You should be able to use the &'static dyn GlobalPacketPool to reduce the 'generics hell' :sweat_smile:  if you want to. This should get rid of L2CAP_MTU and MAX_NOTIF parameters from GattClient.

lulf (approved):
Looks good to me, let me know if you want to address my inline comment or just go with it as is, I'm fine with either, this should be ready to merge once CI passes.

I agree it's not the ideal user experience having to create the packet pool, but maybe one future improvement would be to utilize the 'config system' (see gen_config.py) and host/src/config.rs to parameterize the pool at build time and keep it 'hidden' rather than having to be passed into the client? It limits how flexible you can configure the options, but otoh maybe that's not needed at this level.

host/src/gatt.rs
pferreir: OK, pushed. But I haven't been able to get rid of those parameters. Otherwise, how do I build the corresponding arrays?

lulf: Ah you're right about the notif const. But the L2CAP_MTU should be removable? I don't see it used anywhere. In any case, I think it's fine to merge.

Just rebase it on latest main and we can merge it

pferreir: > Ah you're right about the notif const. But the L2CAP_MTU should be removable? I don't see it used anywhere

it's still used in `request` to build the receive buffer. Anyway, rebased and pushed!

pferreir:
WRT to the test failing, I don't get it how `[u8]` seems to implement `LowerHex` in the pico w but not on the desktop?

lulf:
This is a defmt vs log issue. You can reproduce by compiling trouble-host with the log feature. Basically the format specifiers that works with defmt doesnt always work with core fmt. 

You need to remove that log line or guard it behind a defmt feature gate

pferreir:
Ok, should be fine now!

lulf:
The integration tests fail, you can see the error here https://github.com/embassy-rs/trouble/actions/runs/10798211914/job/29951213402#step:5:153. If you want I can merge it as is and fix the tests after the fact if you're not able to run them yourself (You need 2 HCI dongles). Seems like tests need some increase in the max number of attributes or something.

pferreir:
Oh. How are these tests run in CI, out of curiosity?

lulf:
@pferreir Like this: https://github.com/embassy-rs/trouble/blob/main/.github/workflows/tests.yaml#L30-L37

Basically I have a local GHA with 2 HCI dongles attached which run the tests in https://github.com/embassy-rs/trouble/tree/main/host/tests

PR #112, opened 2024-09-11
Ensure gatt client runs the task

lulf:
Not running the task will prevent the gatt client from processing responses from the gatt server.


lulf:
CC @pferreir

PR #114, opened 2024-09-11
Add attribute to lib.rs to make RustRover happy

okhsunrog:
Without this line RustRover IDE by JetBrains shows error:
```
Unresolved import: `trouble_host::attribute` [E0432]
```
Probably it's the linters issue, but it still would be nice to see green checkbox instead of errors of unresolved imports in the IDE. Can this be merged?

### DIFF
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -30,6 +30,7 @@ pub mod types;
 pub use packet_pool::Qos as PacketQos;
 
 pub mod advertise;
+pub mod attribute;
 pub mod connection;
 pub mod l2cap;
 pub mod scan;


okhsunrog:
Oh, it's already there, just it's optional... Sorry, I didn't look closely at lib.rs before submitting it. Seems like it's RustRover's bug, when I imported `trouble-host` using 
```
trouble-host = { default-features = false, features = ["defmt", "gatt"], git = "https://github.com/embassy-rs/trouble", package = "trouble-host", rev = "d9d2e80b6b2ed6c3aeae1de534de557906353522" }
```
it failed to import modules which were enabled by features and when I removed "package", so my import looked like this:
```
trouble-host = { default-features = false, features = ["defmt", "gatt"], git = "https://github.com/embassy-rs/trouble", rev = "d9d2e80b6b2ed6c3aeae1de534de557906353522" }
```
the import errors were gone. So it's purely RustRovers bug, I'm closing this silly PR

PR #120, opened 2024-09-23
Refactor gatt clients to use feature for consts

lulf:
- Use feature names for notification queue size and max subscribers
- Use embassy-sync pubsub to hold subscribers
- Implement unsubscribe resetting CCCD flag (subscribers can just be drop'ed)
- Document public APIs


lulf:
@pferreir FYI ^^, I think with pubsub we get similar memory requirements, no need to use the packet pool. Also made some of the consts feature flags instead to simplify.

PR #121, opened 2024-09-24
examples: Add support for different devices for nrf-sdc

plaes:
Examples ran on nrf52832, though I wasn't able to connect to l2cap peripheral example using NRF Connect.

Also had to reduce the task arena size to 32k as nrf52832 has only 64k of RAM.


lulf:
I don't think you can use l2cap credit channels with nrf connect, but if you have 2 devkits, you can run the l2cap-central talking to the l2cap-peripheral example

lulf (approved):
Thanks!

Issue #123, opened 2024-09-24
Service discovery request resolves to empty list


ghost:
This issue was identified running the esp32 example on an ESP32-C3-DevKit-RUST-1 board @ 29dabe7 (the esp32 example is currently broken beyond this point, see #122).

Attempts to discover services on the device take a long time to resolve, and when they do they resolve to an empty list.

I have tested this against examples in two different adapter crates running on Ubuntu 20.04:
- [Blteplug](https://github.com/deviceplug/btleplug) (which uses [bluez-async](https://github.com/bluez-rs/bluez-async)) running [this example](https://github.com/deviceplug/btleplug/blob/master/examples/discover_adapters_peripherals.rs)
- [Bluest](https://github.com/alexmoon/bluest) (which uses [bluer](https://github.com/bluez/bluer)) running a combination of code from the [scan](https://github.com/alexmoon/bluest/blob/main/examples/scan.rs) and [connected](https://github.com/alexmoon/bluest/blob/main/examples/connected.rs) examples to scan until the Trouble device is found, connect to it and attempt to discover services

The results from these tests are the same. When the Linux application discover services on the device, I see the following trace from the device:
```
INFO - [adv] connection established
INFO - [gatt] Read event on Characteristic { cccd_handle: Some(35), handle: 34 }        // NOTE: This line doesn't always appear!
ERROR - [gatt] Error processing GATT events: NotFound
INFO - [adv] notifying connection of tick 1
INFO - [adv] notifying connection of tick 2
INFO - [adv] notifying connection of tick 3
(continues)
```

Service UUIDs can be read from the advertising data, but no service data is associated with them.

Running a [bleps](https://github.com/bjoernQ/bleps)-based application on the same dev board, I can successfully discover services with both blteplug and bluest.

Confusingly however, I can connect to the device with the NRFConnect app on my phone and this does seem to be able to resolve the services successfully. I don't know how NRFConnect goes about service discovery, but I don't see either the `Read event on Characteristic { cccd_handle: Some(35), handle: 34 }` line or the `Error processing GATT events: NotFound` line. Instead when NRFConnect connects I just see:
```
INFO - [adv] connection established
INFO - [adv] notifying connection of tick 1
INFO - [adv] notifying connection of tick 2
INFO - [adv] notifying connection of tick 3
INFO - [adv] notifying connection of tick 4
```

lulf:
Several issues are fixed, so we should try to get main branch working on esp32 first.

[...]

PR #127, opened 2024-09-27
Split runner into tx, rx and control

lulf:
@peteskubiak Might be worth trying the esp32 example in this branch. I've made it possible to run the RX task in a separate embassy task, just to check if the connection timeout is due to not polling the HCI adapter frequently enough.


peteskubiak:
I tested this branch with trace enabled for trouble_host but the connection was still slow to establish and no services were found (write up in [this comment](https://github.com/embassy-rs/trouble/issues/123#issuecomment-2382867628)

PR #132, opened 2024-10-06
Create a cargo workspace with testing instructions.

gibbz00:
Not very useful, I know, but I'm interested in contributing to this project further. 

I might send a couple of PRs similar to this one while taking a look around. Feel free to let me know if that's inappropriate :)


lulf:
You can't mix different target architectures in the same workspace, so I'm not sure if this is going to work :/

gibbz00:
Not sure I 100% follow. Are there cases where you are interested in building separate crates for different targets? And why wouldn't the `--target` flag work?

HaoboGu:
It's common for a Rust embedded library to be used in different targets. In this case, cargo workspace won't work. It doesn't support per crate target and per crate features.

gibbz00:
> It's common for a Rust embedded library to be used in different targets. In this case, cargo workspace won't work. It doesn't support per crate target and per crate features.

Sorry, I don't intend to be rude, but you just repeated Ulf without pointing to any definitive sources.

You [can place](https://doc.rust-lang.org/cargo/reference/config.html) `Cargo.toml`s in each crate if workspace commands error due to incorrect target triples. `cargo check` and `cargo test` do already work without any issues, and I've not changed how the ci.sh script invokes each example with an explicit target tripple.

HaoboGu:
> You [can place](https://doc.rust-lang.org/cargo/reference/config.html) Cargo.tomls in each crate if workspace commands error due to incorrect target triples. cargo check and cargo test do already work without any issues, and I've not changed how the ci.sh script invokes each example with an explicit target tripple.

That makes no difference if you use `cargo check -p crate_name` on each sub-crate in workspace, AND it will break rust analyzer, since different crates may require different version of same dependency, which cannot be resolved by cargo right now. Some related discusses can be found [here](https://stackoverflow.com/questions/61273927/target-specific-packages-in-cargo-workspace), [here](https://github.com/rust-lang/cargo/issues/7004) and [here](https://www.reddit.com/r/rust/comments/14x9q0p/comment/jrm96d4/). Monorepo is more preferred way by now.

gibbz00:
My bad, I meant to write `config.toml`.  In fact, that is exactly what the closed issue in the second link you sent recommends. So just to clarify, you aren't referring to an issue cargo but an issue in rust-analyzer? In that case https://github.com/rust-lang/rust-analyzer/issues/10021 is the issue that shold be tracked.

gibbz00:
> since different crates may require different version of same dependency, which cannot be resolved by cargo right now. 

Also, this is very very wrong.

HaoboGu:
> My bad, I meant to write `config.toml`. In fact, that is exactly what the closed issue in the second link you sent recommends. So just to clarify, you aren't referring to an issue cargo but an issue in rust-analyzer? In that case [rust-lang/rust-analyzer#10021](https://github.com/rust-lang/rust-analyzer/issues/10021) is the issue that should be tracked.

The tracking issue is: https://github.com/rust-lang/cargo/issues/9406

gibbz00:
> Involving config.toml requires use cargo check -p crate_name for each crate, which has no advantage comparing with cargo check --manifest-path path.

Awesome, thanks. Now I understand where you're coming from. I tried this in a toy example, and you're right, cargo doesn't seem to build with per package targets.

> The tracking issue is: rust-lang/cargo#9406

Sweet, thank you!

PR #138, opened 2024-10-12
Parse manufacturer data in AdStructure

iamtio:
Implemented parsing of the manufacturer data for AdStructure

### DIFF
--- a/host/src/advertise.rs
+++ b/host/src/advertise.rs
@@ -484,7 +484,10 @@ impl<'d> AdStructureIter<'d> {
             0x08 => Ok(AdStructure::ShortenedLocalName(data)),
             0x09 => Ok(AdStructure::CompleteLocalName(data)),
             // 0x16 => unimplemented!(),
-            // 0xff => unimplemented!(),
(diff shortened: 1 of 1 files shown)


host/src/advertise.rs
```diff
@@ -484,7 +485,10 @@ impl<'d> AdStructureIter<'d> {
             0x08 => Ok(AdStructure::ShortenedLocalName(data)),
             0x09 => Ok(AdStructure::CompleteLocalName(data)),
             // 0x16 => unimplemented!(),
-            // 0xff => unimplemented!(),
+            0xff => Ok(AdStructure::ManufacturerSpecificData {
+                company_identifier: LittleEndian::read_u16(&data[..2]),
```
alexmoon, line 489: You can use `u16::from_le_bytes()` to avoid the `byteorder` dependency. Also, this clause should probably guarded to ensure the data is at least 2 bytes long (e.g. `0xff if data.len() >= 2 => Ok(...)`)

iamtio: Fixed. And thank you. I am new in Rust. Still don't know a lot of things

lulf (approved):

Issue #136, opened 2024-10-10
Consolidate methods and data onto characteristics


petekubiak:
I believe characteristic actions can be simplified by adding methods and persistent data to each characteristic instead of requiring access through the AttributeTable. This has the potential to remove a layer of complexity from the library by removing the need for the AttributeServer and AttributeTable.

This follows on from the feature developed in #131 to define GATT services and characteristics through attribute procedural macros. The idea is to implement a model similar to nrf_softdevice where actions such as `get()`, `set()` and `notify()` are tied to the characteristics themselves rather than being called on the GattServer and AttributeTable types.

This work is closely linked with #135 as both require changes to the same area of the code. I will attempt to put together a draft propsal for this so that we can discuss further how this can be implemented in such a way that it incorporates the Matter BTP requirements outlined in #135.

alexmoon:
That sounds like a pretty good idea to me. I think that should apply generally to Attributes (e.g. either characteristics or descriptors). I also think that they should come in two flavors: host managed or app managed. With a host-managed attribute, the host has the persistent data and updates or it when receiving requests from the remote. With an app-managed attribute, the host uses events/callbacks/whatever to notify the app of remote requests and the app is responsible for providing the response.

lulf:
Agreed it would be nice to use the characteristics directly for this.

PR #141, opened 2024-10-17
Add Nordic Uart Service (NUS) peripheral to examples

plaes:
Add generic peripheral implementation of Nordic Uart Service (NUS) and add example of it to nrf-sdc.

Has still some TODOs

### DIFF
--- a/examples/apps/src/ble_nus_peripheral.rs
+++ b/examples/apps/src/ble_nus_peripheral.rs
@@ -1,5 +1,8 @@
 /* Nordic Uart Service (NUS) peripheral example */
-use embassy_futures::join::join3;
+use embassy_futures::{
+    join::join3,
+    select::{select, Either},
+};
 use embassy_time::{Duration, Timer};
 use heapless::Vec;
 use trouble_host::prelude::*;
@@ -13,15 +16,13 @@ const CONNECTIONS_MAX: usize = 1;
 /// Max number of L2CAP channels.
 const L2CAP_CHANNELS_MAX: usize = 2; // Signal + att
 
-const MAX_ATTRIBUTES: usize = 32;
-
 pub const MTU: usize = 120;
 // Aligned to 4 bytes + 3 bytes for header
 pub const ATT_MTU: usize = MTU + 3;
 
 type Resources<C> = HostResources<C, CONNECTIONS_MAX, L2CAP_CHANNELS_MAX, L2CAP_MTU>;
 
-// GATT Server
+// GATT Server definition
 #[gatt_server]
 struct Server {
     nrf_uart: NrfUartService,
@@ -52,10 +53,11 @@ where
     let server = Server::new_with_config(
         stack,
         GapConfig::Peripheral(PeripheralConfig {
-            name: "TrouBLE",
+            name: "TrouBLE NUS",
             appearance: &appearance::GENERIC_UNKNOWN,
         }),
-    );
+    )
+    .unwrap();
 
     info!("Starting advertising and GATT service");
     let _ = join3(
@@ -70,33 +72,20 @@ async fn ble_task<C: Controller>(mut runner: Runner<'_, C>) -> Result<(), BleHos
     runner.run().await
 }
 
-async fn gatt_task<C: Controller>(server: &Server<'_, C>) {
-    loop {
-        match server.next().await {
-            Ok(GattEvent::Write { handle, connection: _ }) => {
-                let _ = server.get(handle, |value| {
-                    info!("[gatt] Write event on {:?}. Value written: {:?}", handle, value);
-                });
-            }
-            Ok(GattEvent::Read { handle, connection: _ }) => {
-                info!("[gatt] Read event on {:?}", handle);
-            }
-            Err(e) => {
-                error!("[gatt] Error processing GATT events: {:?}", e);
-            }
-        }
-    }
+async fn gatt_task<C: Controller>(server: &Server<'_, '_, C>) -> Result<(), BleHostError<C::Error>> {
+    server.run().await
 }
 
 async fn advertise_task<C: Controller>(
     mut peripheral: Peripheral<'_, C>,
-    server: &Server<'_, C>,
+    server: &Server<'_, '_, C>,
 ) -> Result<(), BleHostError<C::Error>> {
     let mut adv_data = [0; 31];
     AdStructure::encode_slice(
         &[
             AdStructure::Flags(LE_GENERAL_DISCOVERABLE | BR_EDR_NOT_SUPPORTED),
-            AdStructure::CompleteLocalName(b"Trouble NRF UART (NUS) Service"),
+            AdStructure::ServiceUuids16(&[Uuid::Uuid16([0x0f, 0x18])]),
+            AdStructure::CompleteLocalName(b"TrouBLE NUS"),
         ],
         &mut adv_data[..],
     )?;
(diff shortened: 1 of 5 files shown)


examples/apps/src/ble_nus_peripheral.rs
```diff
@@ -0,0 +1,133 @@
...
+    let mut table: AttributeTable<'_, NoopRawMutex, MAX_ATTRIBUTES> = AttributeTable::new();
+
+    // Generic Access Service (mandatory)
+    let id = b"Trouble Example Device";
+    let mut svc = table.add_service(Service::new(0x1800));
+    let _ = svc.add_characteristic_ro(0x2a00, id);
+    svc.build();
+
+    // Generic attribute service (mandatory)
+    table.add_service(Service::new(0x1801));
```
lulf, line 62: Just tagging you here @petekubiak and @jamessizeland, I wonder if it would make sense to have the macros generate these mandatory services? Iirc the old softdevice does that internally.

petekubiak: We discussed this when working on the last PR, and are definitely keen to implement something along these lines at some point.
I'm not an expert in BLE so excuse if this is a silly question, but my understanding is that these services are associated with GAP for a generic BLE device. Is this something we would always want included, or instead an optional feature if the user wants to implement GAP in their application?

lulf: According to the BLE spec (https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-access-profile.html#UUID-8942f34d-80a7-1f7f-8f92-c8ee39aed116_informaltable-idm13358906307630), they are mandatory ("The GATT Server shall contain the GAP service as defined in the GAP Service Requirements in Table 12.1. A device shall have only one instance of the GAP service in the GATT Server. The GAP service is a GATT based service with the service UUID as «GAP Service» defined in [Assigned Numbers](https://www.bluetooth.com/specifications/assigned-numbers).")

So I don't think it should be optional.

petekubiak: Aha, thanks for the info! Turns out I hadn't properly remembered what @jamessizeland had told me - he had pointed out that devices could present as various different types of device depending on the way in which the GAP service was set up. Long story short, this is definitely functionality we'd like to add to the macro yes :)

jamessizeland: Agree we should add it either to the macro or as a builder. I'll sketch out some options tomorrow to see what's most ergonomic.

I'll open an issue to discuss.

jamessizeland: https://github.com/embassy-rs/trouble/issues/142

jamessizeland: This is completed and merged now, should no longer be blocking this for you @plaes :)

examples/apps/src/ble_nus_peripheral.rs
```diff
@@ -0,0 +1,129 @@
...
+                Advertisement::ConnectableScannableUndirected {
+                    adv_data: &adv_data[..],
+                    scan_data: &[],
+                },
+            )
+            .await?;
+        let conn = advertiser.accept().await?;
+
+        /* TODO: Implement "echo" and push rx bytes back to tx? */
+        // XXX: I have to manually "fix" the size from 123 -> 128
```
plaes, line 117: @lulf Any clues what I'm doing wrong here? :)

jamessizeland:
what hardware do I need to try this out?

examples/apps/src/ble_nus_peripheral.rs
petekubiak: @plaes Can you explain a little more about the issue? Does something break if you make the `tx` array 123 bytes?

plaes: If you have following on line 118, it asserts that 123 != 128 when sending the notification after someone has connected:
```rust
let mut tx = [0; ATT_MTU];
```

petekubiak: Interesting...I'm guessing this is the asset in question: https://github.com/embassy-rs/trouble/blob/51d2b95a29b258b7b48f877cdf7af8a0ba54cd75/host/src/attribute.rs#L381

The issue you're facing is because your `tx` characteristic is of type `heapless::Vec<u8, 123>`, while the local variable `tx` is of type `[u8; 123]`. `heapless::Vec` has 4 bytes of metadata associated with it, giving 127 bytes which is probably being padded to 128 for alignment :)

petekubiak: Following up on this, I'd definitely recommend using the same type throughout. With your code as it is currently written, line 124 is actually modifying the first byte in the `heapless::Vec`'s length field (which is of type `usize`). This will cause undefined behaviour if you were to try to use it as a `Vec` elsewhere.

On the whole I would be reluctant to use non-primitive types for characteristics as it assumes that the other end of the comms channel has an understanding of that type. Furthermore, the heapless implementation of Vec could change its memory structure in a future version, which would change the representation of the data in memory and cause further confusion.

TL;DR: in this instance I'd strongly recommend making your `tx` characteristic a byte array.

plaes: Firstly, nice catch regarding the Vec vs byte array mismatch...

Though, using `heapless::Vec` hasn't been issue so far with nrf-softdevice library - https://github.com/lulf/watchful/blob/main/firmware/app/src/ble.rs

petekubiak: I'm not particularly familiar with nrf-softdevice, but I believe that nrf-softdevice has logic to deal with characteristics defined as `Vec` type as variable-length characteristics. So on transmission it will only send `vector.len()` bytes, and on reception it will update the vector's length to the length of the received data.
At present (correct me if I'm wrong @lulf) TrouBLE doesn't have any of this logic and treats all characteristics as a simple block of data. So your `tx` characteristic will be sent as 128 bytes regardless of the current length of the vector.

I think support for variable-length characteristics is certainly something we want to add to TrouBLE though.

plaes: This is where the assertion happens:
https://github.com/embassy-rs/trouble/blob/51d2b95a29b258b7b48f877cdf7af8a0ba54cd75/host/src/attribute.rs#L376-L383

Regarding nrf-softdevice, all the value conversions seem to go through `to_gatt()` and `from_gatt()` methods, which is provided by two traits: [`FixedGattValue`](https://github.com/embassy-rs/nrf-softdevice/blob/991d28e3c61687908acc53e748191eb59d00e27a/nrf-softdevice/src/ble/gatt_traits.rs#L12-L22) and [`GattValue`](https://github.com/embassy-rs/nrf-softdevice/blob/991d28e3c61687908acc53e748191eb59d00e27a/nrf-softdevice/src/ble/gatt_traits.rs#L24-L35).

GattValue can handle data with length smaller than attribute size and also implements support for Vec and String types via Heapless.

jamessizeland: Nice, I suspect we'll want to just implement the same, or possibly do so as part of the bt-hci crate.

petekubiak: Yes, I'm in the process of bringing the FixedGattValue and GattValue traits into TrouBLE as part of my work on #136. The purpose of that work isn't to implement support for variable length characteristics, but it will provide the framework to be able to implement this in future.

petekubiak: I've raised #149 to track the variable length characteristic implementation

lulf: Agreed bringing over FixedGattValue and GattValue traits to trouble makes sense. There is also #133 which might also allow doing varlength without having to store it.

jamessizeland:
are you blocked by anything on finishing this PR that we can help with @plaes?

jamessizeland:
Cool, no worries, I'll see about taking a look!

lulf:
Is this still being worked on, or should it be closed or does anyone want to pick it up?

plaes:
> Is this still being worked on, or should it be closed or does anyone want to pick it up?

I won't have time to deal with this until June.

Issue #142, opened 2024-10-17
Reduce GAP & DIS service boilerplate


jamessizeland:
Proposal that the Generic Access Profile (GAP) and Device Information Service (DIS) should be built into the boilerplate reduction efforts as part of trouble-host-macro crate.

Removing the need for all of the following code:
```rust
    let mut table: AttributeTable<'_, NoopRawMutex, MAX_ATTRIBUTES> = AttributeTable::new();

    // Generic Access Service (mandatory)
    let id = b"Trouble";
    let appearance = [0x80, 0x07];
    let mut svc = table.add_service(Service::new(0x1800));
    let _ = svc.add_characteristic_ro(0x2a00, id);
    let _ = svc.add_characteristic_ro(0x2a01, &appearance[..]);
    svc.build();

    // Generic attribute service (mandatory)
    table.add_service(Service::new(0x1801));
```

The application developer should still be able to configure the appropriate fields, but that they should be made mandatory and checked at compile time.

This could either be with more macro args or as a builder function on-top of Service::new().

Such as 

```rust
let service = Service::new_default(stack, "trouBLE"); 
// Or
let service = Service::new_with_config(stack, gap_config, Some(dis_config));
```

This could also include enums with the standard uuids for i.e. generic ble device, generic mobile, etc.

jamessizeland:
This would actually take out any requirement for table to be external to the Service entirely.

plaes:
In nrf-softdevice, this was handled via "Config" struct:
https://github.com/embassy-rs/nrf-softdevice/blob/991d28e3c61687908acc53e748191eb59d00e27a/nrf-softdevice/src/softdevice.rs#L49-L64

alexmoon:
I'm inclined to think that the Generic Access and Generic Attribute services should be automatically created when you construct the `GattServer` object, since a GATT server is not valid without those services. Handling requests to those services should be handled by the host so it can ensure correct behavior for qualification, with appropriate methods for the app to get/set the information it needs. 

I think the Device Information Service is separate, since it is optional by the spec. It could be useful for `trouble` to provide a ready-to-use implementation of the DIS service, but that could also be provided by an external crate.

jamessizeland:
[...]

PR #143, opened 2024-10-21
Remove gap boilerplate

jamessizeland:
PR to respond to https://github.com/embassy-rs/trouble/issues/142

This is intended to remove additional boilerplate and make setting up a BLE server more accessible for beginners.

As the Generic Access Profile (GAP) service is required, this reduces the user error chance of 
- not creating it
- not understanding what it means
- creating it multiple times

but currently still allows the existing API `let server = server::new(stack, &mut table);` if this is preferred for a reason I'm not aware of.

I've also added some example Appearance UUIDs and a helper function to generate more, as this was something I found confusing to understand.  These UUIDs and others from the spec might be best pulling out into a separate crate in future?

The GATT related modules might also benefit from being reorganized into a module tree in future instead of the current flat hierarchy?

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trouble-host"
-version = "0.1.0"
+version = "0.2.0"
 edition = "2021"
 description = "An async Rust BLE host"
 license = "Apache-2.0 or MIT"
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "trouble-host"
-version = "0.2.0"
+version = "0.1.0"
 edition = "2021"
 description = "An async Rust BLE host"
 license = "Apache-2.0 or MIT"
--- a/host/src/gap.rs
+++ b/host/src/gap.rs
@@ -1,0 +1,65 @@
+//! ## Generic Access Profile
+//!
+//! This profile defines the generic procedures related to discovery of
+//! Bluetooth devices (idle mode procedures) and link management aspects
+//! of connecting to Bluetooth devices (connecting mode procedures).
+//! It also defines procedures related to use of different security levels.
+//! In addition, this profile includes common format requirements for
+//! parameters accessible on the user interface level.
+
+use crate::prelude::*;
+use embassy_sync::blocking_mutex::raw::RawMutex;
+
+/// Configuration for the GAP Service.
+pub enum GapConfig {
+    /// Peripheral device configuration.
+    Peripheral(PeripheralConfig),
+    /// Central device configuration.
+    Central(CentralConfig),
+}
+
+/// Configuration for a peripheral device GAP Service.
+pub struct PeripheralConfig {
+    /// The name of the peripheral device.
+    pub name: &'static str,
+    /// The representation of the external appearance of the device.
+    pub appearance: &'static [u8; 2],
+    // pub preferred_connection_parameters: Option<ConnectionParameters>,
+}
+
+/// Configuration for a central device GAP Service.
+pub struct CentralConfig {
+    /// The name of the central device.
+    pub name: &'static str,
+    /// The representation of the external appearance of the device.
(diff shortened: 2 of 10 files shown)


host/src/gap.rs
```diff
@@ -0,0 +1,122 @@
...
+    /// Peripheral device configuration.
+    Peripheral(PeripheralConfig),
+    /// Central device configuration.
+    Central(CentralConfig),
+}
+
+/// Configuration for a peripheral device GAP Service.
+pub struct PeripheralConfig {
+    /// The name of the peripheral device.
+    pub name: &'static str,
```
lulf, line 65: Would it make sense to 

* Add lifetimes to allow getting values from a stack buffer
* Use &[u8] instead of &str, in cases where you don't want to decode utf-8 in order to pass these values.

lulf (commented):
Nice improvement, I like the general approach!

host/src/gap.rs
alexmoon: >     * Add lifetimes to allow getting values from a stack buffer

Yes, that is important since some devices define their name at runtime. Also, there needs to be a way to change the device name after the server has been started.

>     * Use &[u8] instead of &str, in cases where you don't want to decode utf-8 in order to pass these values.

The spec requires UTF-8 so I think `&str` is appropriate here.

alexmoon:
It might make sense to put the UUIDs in the `bt-hci` crate, along with the other Bluetooth assigned numbers.

We'll also want to create a GATT service along with the GAP service. It defines characteristics like Service Changed, Client Supported Features, Server Supported Features, and DatabaseHash.

host/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "trouble-host"
-version = "0.1.0"
+version = "0.2.0"
```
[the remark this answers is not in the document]
jamessizeland: fair enough, I wasn't sure what the ettiquette was, I'll revert

host/src/gap.rs
jamessizeland: ok I'll see what I can do about the lifetimes, I had some challenges with them and the attribute table.

host/src/gap.rs
```diff
@@ -0,0 +1,122 @@
...
+            }
+            GapConfig::Central(config) => {
+                let id = config.name.as_bytes();
+                let _ = gap.add_characteristic_ro(DEVICE_NAME_UUID, id);
+                let _ = gap.add_characteristic_ro(APPEARANCE_UUID, &config.appearance[..]);
+            }
+        };
+        gap.build();
+
+        table.add_service(Service::new(GATT_UUID)); // GATT UUID (mandatory)
```
jamessizeland, line 120: @alexmoon is this what you mean about the Gatt service? At present its created with no characteristics, as per @lulf's pre-existing boilerplate.

alexmoon: Ah, yes. Sorry, I missed it.

host/src/gap.rs
jamessizeland: I don't think its possible to change the device name after the server has been started in the current implementation. I think to do that we'd need to have it as a read and write characteristic, and then perhaps expose a specific method for changing the id.  That might need to be a future PR.

jamessizeland: I might need a hand with making those lifetimes non-static, I couldn't get the borrow checker to play nice with 'value

lulf: I'm fine merging this as is now. We can address these points later:

* Moving BLE spec numbers to the bt-hci crate (https://github.com/embassy-rs/trouble/issues/145)
* Supporting non-static lifetimes for gap config (https://github.com/embassy-rs/trouble/issues/146)

lulf (approved):
Looks good to me. Once you accept membership to the trouble repo, you should have privileges to run /test as well.

lulf:
Ok, I think it should work now, I had to add you as a collaborator. Please rebase first.

PR #147, opened 2024-10-22
Allow using GapConfig with local lifetime

lulf:
Macro usage still requires static since it uses StaticCell for the table, which requires all values to be 'static as well.

### DIFF
--- a/host-macro/src/server.rs
+++ b/host-macro/src/server.rs
@@ -136,7 +136,7 @@ impl ServerBuilder {
                 /// Create a new Gatt Server instance.
                 ///
                 /// This function will add a GAP Service.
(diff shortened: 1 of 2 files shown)


jamessizeland:
Yeah that helps the first part, the issue of the second part is no owner for the table. I wonder if the GattServer could just take ownership of the table at GattServer::new()

jamessizeland (approved):

lulf:
> Yeah that helps the first part, the issue of the second part is no owner for the table. I wonder if the GattServer could just take ownership of the table at GattServer::new()

Agreed, i think that will simplify the lifetime s

jamessizeland:
happy to close this or do you want us to chase the lifetimes further first?

lulf:
I forgot about it thanks for the reminder :)

PR #151, opened 2024-10-29
Improve characteristic interactions

petekubiak:
This work has evolved considerably from the initial proposal I gave in #136 as my understanding of the code base has grown.

The scope of this PR is to make use of the new macro-style declaration of services and characteristics to improve the server's API. This has led to the following changes:
- Characteristics now hold information on their data type. This is used to infer types for the get(), set() and notify() methods on the server.
- The GattEvent generic callback structure has been changed such that it contains the characteristic's handle rather than the Characteristic type itself. This is due to the Characteristic type holding a generic as mentioned above
- To compensate for this, optional `on_read` and `on_write` callbacks have been implemented for attributes. These will be automatically called when a read or write event occurs on the attribute and have the following signatures:
```rust
fn on_read(connection: &Connection) {}

fn on_write(connection: &Connection, data: &[u8]) -> Result<(), ()> {}  // Where data is the data being written to the attribute
```


petekubiak:
I have another approach which would be to implement the characteristic functions on the server directly rather than on the services. This circumvents all of the issues with referencing and ambiguous types, but has its own set of compromises:
- The gatt_server and gatt_service macros would have to change. There would instead need to be a single macro in which all services and characteristics were declared. This would mean that services could no longer be defined in separate files alongside the functionality associated with them.
- The characteristic functions would no longer be grouped by service. This can be somewhat mitigated by prefixing the characteristic function with the service name (this also eliminates the issue of characteristics with the same name). So rather than `server.service.characteristic_get()` you would have `server.service_characteristic_get()`.

lulf:
> This works for the `_get()` and `_set()` functions, but there is an issue with the `_notify()` function in that the return type is dependent upon the Error type of the Controller used by the GattServer. I've been unsuccessful in finding a way to express this type in the `GattServerInterface` trait and the service's `_notify()` functions which doesn't result in an "ambiguous types" or "mismatched types" error.
> 

Might be worth change the controller-specific error into a different type based on ErrorType::kind() (impemented by Controller traits), so that you can avoid the generics?

In practice, what benefits does this approach give vs. the existing? From the example, it seem like it would change

```
server.notify(server.battery_service.level, &conn, &[tick]).await;
```

to

```
server.battery_service.level_notify(server, &conn, &tick).await;
```

Personally I'm not sure why this is better, but maybe there are better examples to illustrate the advantage of this change.

I would also like to say, that if you need to make significant changes to make the API better, please do not hesitate to do that.

petekubiak:
> Personally I'm not sure why this is better, but maybe there are better examples to illustrate the advantage of this change.
I'm inclined to agree with you. There is still the benefit that these helper functions use the characteristic's type rather than requiring the user to convert to and from byte slices.
I do feel that having to pass in a reference to the server rather muddies the interface though.

> I would also like to say, that if you need to make significant changes to make the API better, please do not hesitate to do that.
I appreciate the endorsement! I'm certainly willing to do this, but I think at present I'm not sure I have a clear enough vision of how it should look.
The most obvious way to implement an interface similar to that of nrf-softdevice would be to ape the structure of nrf-softdevice: that project gets around the issue of needing a reference to the server by having it instantiated as a singleton which can then be acquired as needed in a similar way to peripherals in embedded_hal. We could go down that route, but it's such a fundamental change that I'm wary it may introduce other compromises.
I'd be interested to get @alexmoon 's input on this too, considering his work on nrf-softdevice.

lulf:
Yeah, regarding the use of singletons, I think it should be possible to avoid that, but it does indeed require some experimentation to learn what the ideal structure is. I like the structure of the stuff below GATT at the moment, but maybe for GATT it's worth putting some additional constraints in order to provide a better API.

So maybe the best course of action is to define a few directions for experimenting.

alexmoon:
I agree that singletons tend not to work well in Rust so should be avoided if possible.

It seems like the main virtue of this API is to allow you to use the internal type of the attribute rather than a byte array. Perhaps an alternative like:

```rust
trait AttributeHandle {
  type Value: ToGatt;
  fn raw_handle(&self) -> u16;
}

trait CharacteristicHandle: AttributeHandle {}

struct GattServer {
  ...
  async fn notify<T: CharacteristicHandle>(&self, handle: &T, value: &T::Value) -> Result<...> {
    ...
  }
}
```

> This branch contains another approach in which I have compromised a bit on the new functions' signatures: in this version they take a reference to the macro-generated server as an argument, which circumvents the issue of needing to hold references on the services. The macro-generated server's type can't be accessed directly by the code generated by the `gatt_service` macro, so instead I have created a `GattServerInterface` trait for this purpose.

What if instead the struct generated by the `gatt_server` macro impls `AsRef<trouble::GattServer>`? Then the service methods could be generic on the `AsRef<trouble::GattServer>` bound and you wouldn't have to introduce a new trait.

host/src/gap.rs
```diff
@@ -75,7 +81,7 @@ pub struct PeripheralConfig<'a> {
 /// Configuration for a central device GAP Service.
 pub struct CentralConfig<'a> {
     /// The name of the central device.
-    pub name: &'a [u8; 22],
+    pub name: &'a Vec<u8, DEVICE_NAME_MAX_LENGTH>,
```
alexmoon, line 84: Bluetooth specifies the name to be UTF-8 so I think this should be a `String` instead. And if it is taken by reference it should be a `str` instead of a `String`.

It would also be fine to take it by value as `name: String<DEVICE_NAME_MAX_LENGTH>`, which would eliminate the lifetime parameter. What we can't do is take a `'static` reference like: `name: &'static str` because that makes it difficult to use runtime-determined values for the name.

host/src/gap.rs
```diff
@@ -87,8 +94,11 @@ impl<'a> GapConfig<'a> {
     ///
     /// This configuration will use the `GENERIC_UNKNOWN` appearance.
     pub fn default(name: &'a str) -> Self {
+        static NAME_BYTES: StaticCell<Vec<u8, DEVICE_NAME_MAX_LENGTH>> = StaticCell::new();
+        let name_bytes = NAME_BYTES.init(Vec::new());
+        fill_vec(name_bytes, name.as_bytes());
```
alexmoon, line 99: This will panic if you call `default` more than once.

host/src/attribute.rs
```diff
@@ -393,12 +395,12 @@ impl<'d, M: RawMutex, const MAX: usize> AttributeTable<'d, M, MAX> {
     /// The return value of the closure is returned in this function and is assumed to be infallible.
     ///
     /// If the characteristic for the handle cannot be found, an error is returned.
-    pub fn get<F: FnMut(&[u8]) -> T, T>(&self, handle: Characteristic, mut f: F) -> Result<T, Error> {
+    pub fn get<T: GattValue>(&self, handle: &Characteristic<T>) -> Result<T, Error> {
```
alexmoon, line 398: I think `get` and `set` should operate on Attributes generally, not just Characteristics? You also need to be able to get and set descriptor values and the like.

host/src/gap.rs
petekubiak: Yes this and the static cell below was the result of me running into a sticky situation with  not being able to implement `GattValue` on `&str` because a) it seems to implicitly dereference to `str` which is not `Sized` and b) the lifetime of the associated `&[u8]` gets tied to the lifetime of the `&str`, which according to the compiler doesn't conform to the trait signature.
This code isn't compatible with the API as we want it, so this will definitely be changing.

To get around problem a) I'm going to try implementing a wrapper type around `&str` to see if that avoids the implicit deref.
I think the mem::transmute should take care of problem B) but I need to work on it more.

If you have any ideas about how better to deal with these issues though I'd be very keen to hear them! :)

host/src/gap.rs
petekubiak: Oof yes you're right. I'm planning to replace this implementation anyway as it should be possible to reference the string without having to copy it I think.

host/src/attribute.rs
petekubiak: Oh I didn't realise this! Ok that might need some further thought and development then.

host/src/gap.rs
alexmoon: If the config contains references, you'll want to copy the data into owned storage in the server (e.g. the server will have a buffer somewhere (perhaps in the attributes table) that is `DEVICE_NAME_MAX_LENGTH` bytes and on initialization you'll copy the config data into that buffer). If the config contains owned values, you would just. move the value into the attributes table as part of initialization.

I don't think you need the values in the config struct to be `GattValue`s. The type of the device name characteristic in the attribute table should be `heapless::String<DEVICE_NAME_MAX_LENGTH>` which should be straightforward to implement `GattValue` for.

host/src/attribute.rs
petekubiak: I think you're probably right about this, but I think that as this is the way the attribute table was working before it's worth tackling in a separate PR to avoid this one becoming too large. I'll raise an issue for it.

petekubiak: Raised #155 to implement this.

lulf: Agreed, it can be in a separate PR.

Erik1000:
Hi, just giving you some feedback.
First of, this API is much better than before. I really like the idea of the `GattValue` trait.
Did I understand it correctly that the `on_read` and `on_write` can be some function that can be added to the macro declaration and then will be called on each read write? If so, I find it a good start for a better API. However, one thing I'd like to add: My understanding is that these methods are basically passive and cannot influence the values directly besides instantly overwriting them in some way (really not pretty). I would find it way better if they'd return something like a `Result<T, E>` or `Option<T>` (or perhaps even `ControlFlow`) to signal if the request should either continue with the (potentially changed) value or if it should be dropped. For example, in case a write has invalid data or some client defined application specific requirements, it could cancel the write and therefore the data would not be updated.
Last but not least, I really like the `GattServer::set|get` with the `GattValue`, but the `GattValue` is not used during writes. I think it should be used to parse/verify on write. We can still store the bytes representation in the attribute table (otherwise we will probably get a mess which would need `Box<dyn  GattValue>` or something), but this way, we allow user defined formats and much more customization.

Thank you for your work!

host/src/attribute_server.rs
```diff
@@ -371,19 +369,23 @@ impl<'d, M: RawMutex, const MAX: usize> AttributeServer<'d, M, MAX> {
         Ok(w.len())
     }
 
-    fn handle_read_blob(&self, buf: &mut [u8], handle: u16, offset: u16) -> Result<usize, codec::Error> {
+    fn handle_read_blob(
+        &self,
+        connection: &Connection,
```
jamessizeland, line 374: some places this is conn, others connection

host/src/gap.rs
```diff
@@ -92,31 +92,4 @@ impl<'a> GapConfig<'a> {
             appearance: &appearance::GENERIC_UNKNOWN,
         })
     }
-    /// Add the GAP service to the attribute table.
-    pub fn build<M: RawMutex, const MAX: usize>(self, table: &mut AttributeTable<'a, M, MAX>) {
```
jamessizeland, line 96: There's probably a good reason this is removed and put directly in the macros now. This is a reminder for me to ask you what it is :)

host/src/types/gatt_traits.rs
```diff
@@ -0,0 +1,146 @@
...
+impl Primitive for f64 {}
+
+impl<T: Primitive> FixedGattValue for T {
+    const SIZE: usize = mem::size_of::<Self>();
+
+    fn from_gatt(data: &[u8]) -> Self {
+        if data.len() != Self::SIZE {
+            panic!("Bad len")
+        }
+        unsafe { (data.as_ptr() as *const Self).read_unaligned() }
```
jamessizeland, line 77: lets have a comment here wherever unsafe is required to say why and why it's valid.

host/src/types/gatt_traits.rs
```diff
@@ -0,0 +1,146 @@
...
+    }
+
+    fn to_gatt(&self) -> &[u8] {
+        unsafe { slice::from_raw_parts(self as *const Self as *const u8, Self::SIZE) }
```
jamessizeland, line 81: comment here too

host/src/types/gatt_traits.rs
```diff
@@ -0,0 +1,146 @@
...
+        }
+    }
+}
+
+impl<const N: usize> GattValue for Vec<u8, N> {
+    const MIN_SIZE: usize = 0;
+    const MAX_SIZE: usize = N;
+
+    fn from_gatt(data: &[u8]) -> Self {
+        unwrap!(Self::from_slice(data))
```
jamessizeland, line 105: these unwraps instead of returning a Result, to do with generic challenges from the error type?

jamessizeland (changes requested):
nice work, left a few comments

jamessizeland:
> 
> > Last but not least, I really like the `GattServer::set|get` with the `GattValue`, but the `GattValue` is not used during writes. I think it should be used to parse/verify on write. We can still store the bytes representation in the attribute table (otherwise we will probably get a mess which would need `Box<dyn GattValue>` or something), but this way, we allow user defined formats and much more customization.
> 
> I'm assuming by this you're suggesting that the signature of the write callback takes `data` as type `&impl GattValue` so that you can define an attribute's write callback with the appropriate concrete type? I agree that it would be really nice and clean if the data type can be inferred automatically for the callback signature and I will certainly give this a try. I'm not sure whether it will allow this though because the callback is stored as a function pointer on the attribute and I don't know whether a function pointer type can contain an `impl <Trait>` argument type. If it doesn't work however, there is always the option of calling `T::from_gatt(data)` within your callback in order to convert the byte slice into the appropriate type :)

The benefits of the current way, having it as raw bytes, is that if the incoming data is incorrect its possible to debug.  And you could feasibly have multiple characteristics register the same callback, which would make the signature morphing even harder?

host-macro/src/server.rs
```diff
@@ -119,40 +125,68 @@ impl ServerBuilder {
...
                 ///
                 /// This function will add a Generic GAP Service with the given name.
-                #visibility fn new_default(stack: Stack<'reference, C>, name: &'values str) -> Self {
+                /// The maximum length which the name can be is 22 bytes (limited by the size of the advertising packet).
+                /// If a name longer than this is passed, Err() is returned.
+                #visibility fn new_default(stack: Stack<'reference, C>, name: &'values str) -> Result<Self, &'static str> {
                     let mut table: AttributeTable<'_, #mutex_type, #attribute_data_size> = AttributeTable::new();
 
-                    GapConfig::default(name).build(&mut table);
+                    static DEVICE_NAME: static_cell::StaticCell<HeaplessString<{DEVICE_NAME_MAX_LENGTH}>> = static_cell::StaticCell::new();
```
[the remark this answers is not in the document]
petekubiak: The reason I didn't do that was because if the user attempted to create a second GapConfig then it would panic. Admittedly this implementation would panic if the user attempted to create a second instance of the type decorated with `#[gatt_server]`, but that would happen anyway due to the characteristic data stores.
If a user is only ever going to create one GapConfig then moving this code to a GapConfig method removes some duplicate code, which I'm always for.

host/src/attribute_server.rs
petekubiak: Yeah the places with `connection` are the ones I added. I'll align them all for neatness.

host/src/types/gatt_traits.rs
petekubiak: I have a feeling I went round this particular loop at one point, but I can't remember if I encountered an issue generics or whether I was in pursuit of something else and simply forgot about the unwraps. I'll take another look.

petekubiak:
> The benefits of the current way, having it as raw bytes, is that if the incoming data is incorrect its possible to debug. And you could feasibly have multiple characteristics register the same callback, which would make the signature morphing even harder?

I was thinking it ought to be possible to add some trace to the calling code to report if incorrect data was received and couldn't be converted into the expected type.
You raise a good point about potentially using the same callback for multiple attributes - requiring the callback signature to conform to the attribute's type would definitely make using the same callback for multiple characteristics impossible unless they were all the same type.

host/src/types/gatt_traits.rs
ScottGibb: As well as is the reason we are using the macro version, just for help debugging in the case of a panic?
https://github.com/canndrew/unwrap

host/src/types/gatt_traits.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn to_gatt(&self) -> &[u8] {
+        self.as_slice()
+    }
+}
+
+impl<const N: usize> GattValue for String<N> {
+    const MIN_SIZE: usize = 0;
+    const MAX_SIZE: usize = N;
+
+    fn from_gatt(data: &[u8]) -> Self {
```
ScottGibb, line 136: Is it possible to bubble up this error?
Im guessing bubbling generic errors is quite difficult?/ an absolute pain

host/tests/gatt.rs
```diff
@@ -41,20 +41,20 @@ async fn gatt_client_server() {
         // Random starting value to 'prove' the incremented value is correct
         let mut value: [u8; 1] = [rand::prelude::random(); 1];
         let mut expected = value[0].wrapping_add(1);
-        let mut svc = table.add_service(Service::new(0x1800));
-        let _ = svc.add_characteristic_ro(0x2a00, id);
-        let _ = svc.add_characteristic_ro(0x2a01, &appearance[..]);
+        let mut svc = table.add_service(Service::new(0x1800), None);
+        let _ = svc.add_characteristic_ro(0x2a00, id, None);
+        let _ = svc.add_characteristic_ro(0x2a01, &appearance, None);
```
ScottGibb, line 46: can this be written as
```
let mut svc = table.add_service(Service::new(0x1800), None);
svc.add_characteristic_ro(0x2a00, id, None);
svc.add_characteristic_ro(0x2a01, &appearance, None);
```
Similar to how these Builder patterns look?
https://www.shuttle.dev/blog/2022/06/09/the-builder-pattern

host/tests/gatt.rs
```diff
@@ -196,23 +196,23 @@ async fn gatt_client_server() {
             (Err(e1), Err(e2)) => {
                 println!("Central error: {:?}", e1);
                 println!("Peripheral error: {:?}", e2);
-                assert!(false);
+                panic!();
```
ScottGibb, line 199: Ahhhhhhh!!

ScottGibb (commented):
Looks really good @petekubiak!

host-macro/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "trouble-host-macros"
-version = "0.1.0"
+version = "0.2.0"
```
lulf, line 3: No need to bump it yet, it's not published.

host/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "trouble-host"
-version = "0.1.0"
+version = "0.2.0"
```
lulf, line 3: No need to bump

host/tests/gatt.rs
```diff
@@ -41,20 +41,20 @@ async fn gatt_client_server() {
...
+        table.add_service(Service::new(0x1801), None);
 
         // Custom service
-        let value_handle = table.add_service(Service::new(SERVICE_UUID.clone()))
+        let value_handle = table.add_service(Service::new(SERVICE_UUID.clone()), None)
             .add_characteristic(
                 VALUE_UUID.clone(),
                 &[CharacteristicProp::Read, CharacteristicProp::Write, CharacteristicProp::Notify],
-                &mut value)
+                &mut value, None, None)
```
lulf, line 57: It's a bit annoying that these optional parameters creep up in the API. I think it would be better to use some kind of builder pattern to avoid that.

lulf (commented):
Looks great, I really like the value conversion stuff. My only gripe is those optional parameters for the callbacks.

It also feels like on_read and on_write give you the same thing as processing events with server.next(), except them being non-async. Maybe we can find a way to unify it in the future?

host/Cargo.toml
```diff
@@ -23,6 +23,8 @@ static_cell = "2.1.0"
 # Logging
 log = { version = "0.4.16", optional = true }
 defmt = { version = "0.3", optional = true }
+num_enum = { version = "0.7.3", default-features = false }
+thiserror = { version = "2.0.0", default-features = false }
```
lulf, line 27: @petekubiak Just to be sure, can you measure code size when enabling thiserror? I'm worried it adds some bloat.

petekubiak: I checked out this commit which added `thiserror` (9f7bfc4) and ran `cargo size --release` on the esp32 example. I then did the same thing for the previous commit (000d1e1). Here are the results:
|| 000d1e1 | 9f7bfc4 |
|--------|--------|--------|
|text| 575266 | 575314 |
|data| 1144 | 1144 |
|bss| 144244 | 144244 |
|dec| 720654 | 720702 |
|hex| aff0e | aff3e | 

This would indicate that the inclusion of thiserror has only added 48 bytes, but I guess that only the code that's being used is being included in the binary.
Are there any other ways you'd suggest I should check this?

host/src/types/gatt_traits.rs
petekubiak: It's because `read_unaligned` is an unsafe function and must therefore be inside an unsafe block

host/Cargo.toml
lulf: No, that would be the same way I'd check it. Could you check with the nRF as well, which uses defmt. Increase in size when using log is expected. I'm a bit concerned if this adds to binary size if defmt is used.

host/tests/gatt.rs
ScottGibb: Ahh nice

host/Cargo.toml
petekubiak: Running `cargo size --bin ble_bas_peripheral --release --features nrf52833` on the same two commits gives:
198820 bytes without `thiserror`
198840 bytes with `thiserror`

So only a change of 20 bytes in this case :)

host/src/att.rs
```diff
@@ -28,70 +31,52 @@ pub(crate) const ATT_READ_BLOB_REQ: u8 = 0x0c;
...
 
+/// Attribute Error Code
+///
+/// This enum type describes the `ATT_ERROR_RSP` PDU from the Bluetooth Core Specification
+/// Version 6.0 | Vol 3, Part F (page 1491)
 #[cfg_attr(feature = "defmt", derive(defmt::Format))]
-#[derive(Debug, PartialEq, Clone, Copy)]
+#[derive(Debug, PartialEq, Clone, Copy, TryFromPrimitive, Error)]
 #[repr(u8)]
 pub enum AttErrorCode {
```
alexmoon, line 41: I think we should consider making this a newtype wrapping `u8` with associated constants, rather than an enum. When a client sends a request to a remote server that server can theoretically return any error code it wants, regardless of the spec. For example, if it's a server that supports a newer version of the spec than trouble it could return a code unknown to trouble.

host-macro/src/server.rs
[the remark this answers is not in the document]
petekubiak: a6d1738830df4dd00c8493e1d542f04ec7d17d00

host/src/attribute_server.rs
petekubiak: 95775811e30216eb1518a0688fe5f81e527f22f2

host/src/types/gatt_traits.rs
petekubiak: 39ed273be230f81c790b6ccb0e10145e99d46722

host/src/types/gatt_traits.rs
petekubiak: 39ed273be230f81c790b6ccb0e10145e99d46722

host/src/types/gatt_traits.rs
petekubiak: Turns out I got distracted somewhere along the way. `from_gatt` is now fallible c924901f9f2bb5fd618367917ac8c1ebd934cd13

host/src/types/gatt_traits.rs
petekubiak: c924901f9f2bb5fd618367917ac8c1ebd934cd13

host-macro/Cargo.toml
petekubiak: 23b13fb11d984f4f5b5ae40adab18690d424de47

host/Cargo.toml
petekubiak: Reverted in 23b13fb11d984f4f5b5ae40adab18690d424de47

host/src/att.rs
petekubiak: You make a good point, there is actually a range reserved for application-defined error codes which this wouldn't be able to handle. I think it's beyond the scope of this PR as the logic here hasn't changed, but I'll raise it as an issue to be addressed.

petekubiak:
> It also feels like on_read and on_write give you the same thing as processing events with server.next(), except them being non-async. Maybe we can find a way to unify it in the future?

The main difference between the callbacks and server.next() is that the callbacks trigger before the event is processed. Hopefully the new signature of the write callback allowing the application code a say in accepting / rejecting the request helps to differentiate it and give it more of a unique purpose.

I did wonder whether there were any other parameters we might want to provide to the callbacks. If anyone has any ideas I'd be keen to hear them.

I did originally try to implement them as async, but I can't use the `&impl Future<Output=T>` return type for a function pointer, so would have had to resort to `Box` to be able to make them async.

host-macro/src/server.rs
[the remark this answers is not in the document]
petekubiak: Ah interesting - we were discussing this earlier and thought that creating multiple `#[gatt_servers]` in the same application wouldn't be required.

In that case I will need to revert a6d1738 as it will not allow for a second `GapConfig::Peripheral` to be made (whereas previously you could make as many `GapConfigs` as you desired but could only create one instance of a particular `#[gatt_server]` (but declaring a second `#[gatt_server]` and instantiating that would work fine).

I notice you said "putting an owned `heapless::String` directly into the `GapConfig`" rather than using a statically allocated `heapless::String` - the reason I haven't done that is that the `GapConfig` only exists for a short period of time, so the user would need to hang on to the `GapConfig` for the lifetime of the `#[gatt_server]`.

I originally tried passing ownership of the `GapConfig` to the `#[gatt_server]`, but this ends up in a chicken-and-egg situation where a reference to the `String` in the `GapConfig` is required for the attribute, but the `GapConfig` needs to be moved into the `#[gatt_server]` to complete its initialisation, thus invalidating the reference. Moving the `GapConfig` into the `#[gatt_server]` first and then adding the attributes has the same problem, as the `#[gatt_server]` is returned from the function at the end of initialisation, again causing a move and invalidating the reference.

I encountered this issue a few times during the course of this work - is there a pattern to get around it without using statics?

petekubiak:
> Last but not least, I really like the `GattServer::set|get` with the `GattValue`, but the `GattValue` is not used during writes. I think it should be used to parse/verify on write. We can still store the bytes representation in the attribute table (otherwise we will probably get a mess which would need `Box<dyn GattValue>` or something), but this way, we allow user defined formats and much more customization.

Just a quick update to say that I tried this but `&impl Trait` is unfortunately not allowed in function pointers, so it is not possible to implement the write callback to take the type expected for the callback's attribute.
As I mentioned earlier though, you can still convert the byte slice into the correct data type within your callback using the `<type>::from_gatt()` function.
Please note that `from_gatt()` now returns a Result type, the variant of which is dependent upon length-checking of the slice and any other fallible operations which may be required to complete the conversion.
(I've just realised there's still a panic in the implementation of `FixedGattValue` for `T: Primitive`. I'll remove this tomorrow!)

lulf:
> /test

Agent is back up now.

host/Cargo.toml
lulf: Ok, that is probably this line https://github.com/embassy-rs/trouble/blob/main/host/src/attribute.rs#L268  - which can be fixed separately.

Thanks for checking!

host/src/att.rs
jamessizeland: https://github.com/embassy-rs/trouble/issues/161

jamessizeland (approved):

host/src/attribute.rs
```diff
@@ -367,19 +409,42 @@ impl<'d, M: RawMutex, const MAX: usize> AttributeTable<'d, M, MAX> {
...
-    pub fn set(&self, handle: Characteristic, input: &[u8]) -> Result<(), Error> {
+    pub fn set<T: GattValue>(&self, handle: &Characteristic<T>, input: &T) -> Result<(), Error> {
+        let gatt_value = input.to_gatt();
         self.iterate(|mut it| {
             while let Some(att) = it.next() {
                 if att.handle == handle.handle {
                     if let AttributeData::Data { props, value } = &mut att.data {
-                        assert_eq!(value.len(), input.len());
-                        value.copy_from_slice(input);
+                        assert_eq!(value.len(), gatt_value.len());
```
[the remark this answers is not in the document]
petekubiak: 394cba92ae5d074c0b798e3641176dd865b5d89a

Erik1000:
@petekubiak I just switched my application to your branch! One thing I noticed is that `GattServer::notify` takes `handle: Characteristic<T>` but I think only a `&Charateristic<T>` is required for `self.server.table.set(&handle, value)?;`. Can we change that? Because it is annoying if the type `T` does not implement `Copy` and avoids potentially expensive cloning.

petekubiak:
> @petekubiak I just switched my application to your branch! One thing I noticed is that `GattServer::notify` takes `handle: Characteristic<T>` but I think only a `&Charateristic<T>` is required for `self.server.table.set(&handle, value)?;`. Can we change that? Because it is annoying if the type `T` does not implement `Copy` and avoids potentially expensive cloning.

Thanks for this, real-world testing is always appreciated! Yes, seems like that change can be made without changing anything else in the code - good spot!
https://github.com/embassy-rs/trouble/pull/151/commits/adb1c9b52c3d8ddbaf915406279decaec0a359e7

lulf (approved):

lulf (approved):

petekubiak:
> It also feels like on_read and on_write give you the same thing as processing events with server.next(), except them being non-async. Maybe we can find a way to unify it in the future?

I do think that being able to use async in the callbacks would provide huge benefit. I was talking to @jamessizeland about this yesterday and we realised that `heapless::Box` exists, which would enable us to store function pointers with a `Future` return type.
This may incur a performance penalty due to the memory management and it requires setting up a memory pool for `heapless::Box` to use, so I'm thinking it would be an optional feature.
I'll create an issue for it.

lulf:
If you require async, you could also use the server.next() and handle it there. I think with Box you'd anyway end up needing to store the value first for the future to reach it.

PR #170, opened 2024-11-15
Add ability to receive connection events

lulf:
Add a channel per connection for receiving events related to a connection. At the moment only the Disconnected event is in use, but it can be improved to support GATT events as well.


alexmoon:
Can we add the disconnect reason to the Disconnected event? It's sometimes important to know why the connection was terminated.

lulf:
> Can we add the disconnect reason to the Disconnected event? It's sometimes important to know why the connection was terminated.

I have added it now, using the bt_hci Status type (the reason it can't use the DisconnectedReason enum is that a HCI controller may emit values outside the bounds of the DisconnectedReason enum, which describes valid input to the Disconnect HCI command, not necessarily values returned by the DisconnectedEvent).

jamessizeland (approved):
Looks good! Disconnect works as I imagined.  You can see the minimal diff here: https://github.com/jamessizeland/microbit-ble-gamepad/pull/3/files#diff-553ad6fdf019ddb8a60db5ed3806e406ebaf52a6adb78435b589790255d3df09L67-L70

jamessizeland:
This has made me think we can improve the way GattEvent::Write is parsed.  Its not very ergonomic to have to go

```rust
Ok(GattEvent::Write { value_handle, .. }) => {
    let characteristic = server.server().table().find_characteristic_by_value_handle(value_handle).unwrap();
    let value = server.get(&characteristic).unwrap();
}
```

best would be being able to match on all available Characteristics tagged as Write like:

```rust
Ok(GattEvent::Write { characteristic, .. }) => {
    match characteristic {
        Mode(val) => info!("Switching to mode: {}", val),
        Player(val) => info("Setting controller to player: {}", val),
    }
}
```

host/src/connection.rs
```diff
@@ -31,6 +31,20 @@ pub struct ConnectParams {
...
+pub enum ConnectionEvent<'d> {
+    /// Connection disconnected.
+    Disconnected {
+        /// The reason (status code) for the disconnect.
+        reason: Status,
+    },
+    #[allow(unused, missing_docs)]
+    /// GATT write event.
+    // TODO: Might need to wrap this in a type that holds a Pdu<'d> or a resources from a pool
+    GattWrite { data: &'d [u8] },
```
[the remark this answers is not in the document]
lulf: Yeah, I can remove it. I had originally put it in there so that I could wire the lifetimes of the ConnectionEvent correctly, since that would be required for events that reference Pdu's that were allocated.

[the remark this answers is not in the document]
lulf: Maybe it can stay for now, since it will probably make life easier for anyone making the change to provide gatt write events?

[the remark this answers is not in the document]
jamessizeland: Yep, perhaps just a docstring so an API user knows not to expect any data from it yet?

lulf:
I modified the PR to post the gatt events on the connection events instead of having them returned by the server. The GATT server now only requires run() to be called in some async task, and it will do the dispatching to the connections.


> This has made me think we can improve the way GattEvent::Write is parsed. Its not very ergonomic to have to go
> 
> ```rust
> Ok(GattEvent::Write { value_handle, .. }) => {
>     let characteristic = server.server().table().find_characteristic_by_value_handle(value_handle).unwrap();
>     let value = server.get(&characteristic).unwrap();
> }
> ```
> 
> best would be being able to match on all available Characteristics tagged as Write like:
> 
> ```rust
> Ok(GattEvent::Write { characteristic, .. }) => {
>     match characteristic {
>         Mode(val) => info!("Switching to mode: {}", val),
>         Player(val) => info("Setting controller to player: {}", val),
>     }
> }
> ```


I agree with that, however it would require some sort of TX/GATT packet pool for dispatching the event to the connection and/or sending the replies. At present, the att reply is encoded within the pdu that was received, to save the extra allocation. I think the ergonomics might be worth it though, but I'd like to merge this PR first.

host/src/lib.rs
```diff
@@ -351,6 +354,8 @@ pub fn new<
 
     let connections = &mut *resources.connections.write([ConnectionStorage::DISCONNECTED; CONNS]);
     let connections = unsafe { transmute_slice(connections) };
+    let events = &mut *resources.events.write([const { EventChannel::new() }; CONNS]);
+    let events = unsafe { transmute_slice(events) };
```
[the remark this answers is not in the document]
lulf: It is already there (though not visible in the diff): https://github.com/embassy-rs/trouble/blob/acb4a8f620383108c7d641f4caa675af11387436/host/src/lib.rs#L344-L348

host/tests/gatt_derive.rs
```diff
@@ -137,11 +108,34 @@ async fn gatt_client_server() {
...
+                                break;
+                            }
+                            ConnectionEvent::Gatt { connection: _, event } => match event {
+                                GattEvent::Write {
+                                    value_handle: handle
+                                } => {
+                                    let characteristic = server.server().table().find_characteristic_by_value_handle(handle).unwrap();
+                                    let value: u8 = server.server().table().get(&characteristic).unwrap();
+                                    assert_eq!(expected, value);
+                                        expected = expected.wrapping_add(1);
```
petekubiak, line 127: I think you'll need to add 2 here - in gatt_derive the first write is rejected by the on_write callback of the characteristic (to check the mechanism is working)

petekubiak (approved):
LGTM!

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -119,13 +104,22 @@ async fn advertise_task<C: Controller>(
...
-            info!("[adv] notifying connection of tick {}", tick);
-            let _ = server.notify(&server.battery_service.level, &conn, &tick).await;
+        loop {
+            match select(conn.next(), Timer::after(Duration::from_secs(2))).await {
+                Either::First(event) => match event {
+                    ConnectionEvent::Disconnected { reason } => {
+                        info!("[adv] disconnected: {:?}", reason);
+                        break;
+                    }
+                    _ => {}
```
jamessizeland, line 115: https://github.com/jamessizeland/microbit-ble-gamepad/pull/3/files# updated my PR to show how I'd use this new API design. Looks good! My only suggestion would be to make it { event, connection }, so we can { event, .. } out the connection when not used? (I'm assuming event is the more commonly used return variable.)

```rust
pub async fn gatt_server_task(server: &GamepadServer<'_>, conn: &Connection<'_>) {
    // check if the connection is still active every second
    loop {
        if let Either::First(event) = select(conn.next(), server.run()).await {
            match event {
                ConnectionEvent::Disconnected { reason } => {
                    info!("[gatt] Disconnected: {:?}", reason);
                    break;
                }
                ConnectionEvent::Gatt {
                    connection: _,
                    event,
                } => match event {
                    GattEvent::Read { value_handle } => {
                        info!("[gatt] Server Write event on {:?}", value_handle);
                    }
                    GattEvent::Write { value_handle } => {
                        info!("[gatt] Read event on {:?}", value_handle);
                    }
                },
            }
        }
    }
    info!("Gatt server task finished");
}
```

host/src/connection_manager.rs
```diff
@@ -80,10 +80,14 @@ impl<'d> ConnectionManager<'d> {
         })
     }
 
-    pub(crate) async fn event(&self, index: u8) -> ConnectionEvent<'_> {
+    pub(crate) async fn next(&self, index: u8) -> ConnectionEvent<'d> {
```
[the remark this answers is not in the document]
lulf: According to rustc, yes. I'm not entirely sure why it didn't complain about it before.

host/tests/gatt_derive.rs
lulf: Ah, thanks, I blindly copied this from the other test.

examples/apps/src/ble_bas_peripheral.rs
lulf: You can do { event, ..} independently of the order in the struct. My personal preference is to keep them all in the tests and examples to clearly show what data you get from looking at the code.

host/tests/gatt_derive.rs
lulf: Hmm, I'm not sure if it's correct to change it, the old test was already doing a +=1

lulf: Apologies, you are correct, I was looking at the writes, not expected :man_facepalming:

PR #171, opened 2024-11-21
Change server macro to accept expressions for attribute_data_size and mtu

petekubiak:
Previously, the `server` attribute macro would only accept literals for the attribute_data_size and mtu arguments. This meant that constants could not be used as they are evaluated after the macro has been expanded.

This PR changes the arguments to accept expressions, so constants can now be used.

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -18,8 +18,10 @@ const MAX_ATTRIBUTES: usize = 10;
 
 type Resources<C> = HostResources<C, CONNECTIONS_MAX, L2CAP_CHANNELS_MAX, L2CAP_MTU>;
 
+const ATTRIBUTE_DATA_SIZE: usize = 10;
+
 // GATT Server definition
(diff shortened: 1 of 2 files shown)


jamessizeland (approved):

lulf (approved):
LGTM. One thought: the attribute_data_size should be possible to derive inside the macro right? (It knows how many attributes it will require).

petekubiak:
> LGTM. One thought: the attribute_data_size should be possible to derive inside the macro right? (It knows how many attributes it will require).

This is a problem I've puzzled over before: the number of attributes needs to be known inside the `gatt_server` macro, but the attributes themselves are defined in the `gatt_service` macro - we would need the `gatt_service` macro to make the information on the number of attributes it contains available to be referenced in the `gatt_server`. I think I can do this by defining a `const ATTRIBUTE_COUNT` within the `impl` of each service, then generating code for the server to automatically sum these up.

I suggest keeping the optional macro argument as an override in case the user wants to leave some extra room in the table to manually add more attributes.

PR #174, opened 2024-11-21
Improve peripheral gatt example

jamessizeland:
Happy to debate the restructure of the Gatt Peripheral example.  To me this represents a more real world example where separate services will be talking to the gatt server, and can be shut down when the connection drops.

I've added a couple of quality of life improvements alongside this.

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -1,4 +1,4 @@
-use embassy_futures::select::{select, Either};
+use embassy_futures::select::select;
 use embassy_time::Timer;
 use trouble_host::prelude::*;
 
@@ -44,9 +44,8 @@ pub async fn run<C>(controller: C)
 where
     C: Controller,
 {
-
-    // Using a fixed seed means the "random" address will be the same every time the program runs, 
-    // which can be useful for testing. If truly random addresses are required, a different, 
+    // Using a fixed seed means the "random" address will be the same every time the program runs,
+    // which can be useful for testing. If truly random addresses are required, a different,
     // dynamically generated seed should be used.
     let address = Address::random([0x41, 0x5A, 0xE3, 0x1E, 0x83, 0xE7]);
     info!("Our address = {:?}", address);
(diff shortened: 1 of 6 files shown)


examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -63,80 +66,125 @@ where
...
 
-    info!("Starting advertising and GATT service");
-    let _ = join3(
-        ble_task(runner),
-        gatt_task(&server),
-        advertise_task(peripheral, &server),
-    )
-    .await;
+    let ble_runner_task = ble_task(runner);
+    let app_task = async {
```
ScottGibb, line 70: So is this app task here so that whenever a connection is dropped we go back to advertising? Including if there is an error for somer reason within the advertise function?

jamessizeland: app_task here is created so that we can run the main application loop alongside the ble_runner_task. The runner task is required to be running in the background whilst we advertise, connect, read, write, etc on the ble server, otherwise it hangs. 

In our code we've made ble_runner_task into a statically defined #[embassy::task], so it can be spawned in the background and then we can move on to the application logic. 

In this case however we can't do that because the runner is defined with a generic Controller and embassy tasks cannot contain generics.

So we have to await the ble_runner_task but also run our application logic simultaneously, and hence use that last select statement.

petekubiak (approved):

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -63,80 +66,125 @@ where
...
-    mut peripheral: Peripheral<'_, C>,
+/// Stream Events until the connection closes.
+async fn gatt_task<C: Controller>(
     server: &Server<'_, '_, C>,
+    conn: &Connection<'_>,
 ) -> Result<(), BleHostError<C::Error>> {
-    let mut adv_data = [0; 31];
+    let level = server.battery_service.level;
+    loop {
+        if let Either::First(event) = select(conn.next(), server.run()).await {
```
lulf, line 114: I personally think it is a bit counter intuitive to handle both the generic 'run()' task and the connection events in the same task. In a real world situation with multiple connections, this code wouldn't be correct, since you'd want to have only 1 instance of server.run(), and multiple tasks doing conn.next().

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -63,80 +66,125 @@ where
...
+        )
+        .await?;
+    info!("[adv] advertising");
+    let conn = advertiser.accept().await?;
+    info!("[adv] connection established");
+    Ok(conn)
+}
+
+/// Example task to use the BLE notifier interface.
+async fn example_application_task<C: Controller>(server: &Server<'_, '_, C>, conn: &Connection<'_>) {
```
lulf, line 178: I think this is the task that should do all processing of the connection, maybe also rename it connection_task or something?

jamessizeland: I wanted to separate out the connection from this example task, as you could have several tasks all associated with the same connection. (i.e. one for each sensor on your system such as the example below from my Gamepad demo)

```rust
    // Main loop
    loop {
        display.display(QuestionMark, Duration::from_secs(2)).await;
        // advertise for connections
        if let Ok(conn) = advertiser.advertise().await {
            let pause = Duration::from_secs(1);
            speaker.play_tune(Tune::Connect).await;
            display.display_blocking(Heart, pause).await;


            let gatt = gatt_server_task(server, &conn);
            let buttons = buttons_task(&mut gamepad_buttons, &conn, &display);
            let analog = analog_stick_task(server, &conn, &mut analog_stick, &display);
            embassy_futures::select::select3(gatt, buttons, analog).await;
            speaker.play_tune(Tune::Disconnect).await;
        }
```

examples/apps/src/ble_bas_peripheral.rs
jamessizeland: makes sense, I've split them out again

lulf (approved):
Looks good! Maybe example_application_task could be renamed to something more descriptive of what it's doing, maybe counting_task or something.

PR #176, opened 2024-11-22
Automatically calculate attribute table size

### LINKED ISSUE #173: Enhance gatt_server macro to automatically size the attribute table based on the number of defined attributes
> LGTM. One thought: the attribute_data_size should be possible to derive inside the macro right? (It knows how many attributes it will require).

I really like this idea.

This is a problem I've puzzled over before: the number of attributes needs to be known inside the `gatt_server` macro, but the attributes themselves are defined in the `gatt_service` macro - we would need the `gatt_service` macro to make the information on the number of attributes it contains available to be referenced in the `gatt_server`. I think I can do this by defining a `const ATTRIBUTE_COUNT` within the `impl` of each service, then generating code for the server to automatically sum these up.

I suggest keeping the optional macro argument as an override in case the user wants to leave some extra room in the table to manually add more attributes.

_Originally posted by @petekubiak in https://github.com/embassy-rs/trouble/issues/171#issuecomment-2491366007_

petekubiak:
Closes #173.

This PR implements automatic sizing of the attribute table when using the `gatt_server` macro.
The number of attributes for each service is calculated in the `gatt_service` macro, and this is then summed up by the `gatt_server` macro to determine the size of the table.
The option to override this and manually set the attribute table size still exists.
I have also renamed the attribute macro argument to `attribute_table_size` as this more accurately reflects the value being set.

## Topic for discussion
I have implemented a static assertion to throw a compile-time error if the table size is overridden with a value too small to hold the calculated number of attributes. The advantage of this is of course that the user will be notified before flashing their device, meaning that they do not need to be monitoring the device's trace to see the issue.
However, I cannot use formatted strings in a "constant assertion", so the compile-time error looks like this:
```
error[E0080]: evaluation of constant value failed
  --> src/ble_bas_peripheral.rs:22:1
   |
22 | #[gatt_server(attribute_table_size = 9)]
   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the evaluated program panicked at 'Specified attribute table size is insufficient. Please increase attribute_table_size or remove the argument entirely to allow automatic sizing of the attribute table.', src/ble_bas_peripheral.rs:22:1
   |
   = note: this error originates in the macro `$crate::panic::panic_2021` which comes from the expansion of the macro `core::assert` (in Nightly builds, run with -Z macro-backtrace for more info)
```

The alternative would be to use a normal `assert!` which would be evaluated at run-time. This would allow a formatted message of something like:
```
panicked at /home/pkubiak/repos/trouble/examples/apps/src/ble_bas_peripheral.rs:22:1:
Specified attribute table size of 9 is insufficient. A table size of at least 10 is required for the current GATT server configuration.
```
but of course this would only be visible in the device's trace.

I personally lean toward the static assertion, but I'd be keen to hear other opinions on this!

### DIFF
--- a/host-macros/src/server.rs
+++ b/host-macros/src/server.rs
@@ -4,13 +4,10 @@
 //! It should contain one or more Gatt Services, which are decorated with the `#[gatt_service(uuid = "...")]` attribute.
 
 use darling::Error;
-use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::{quote, quote_spanned};
 use syn::{meta::ParseNestedMeta, parse_quote, spanned::Spanned, Expr, Result};
 
-/// Default size for the memory block storing attribute data in bytes
-const DEFAULT_ATTRIBUTE_DATA_SIZE: usize = 32;
 /// MTU for a legacy BLE packet
 const LEGACY_BLE_MTU: usize = 27;
 
--- a/host-macros/src/server.rs
+++ b/host-macros/src/server.rs
@@ -4,6 +4,7 @@
 //! It should contain one or more Gatt Services, which are decorated with the `#[gatt_service(uuid = "...")]` attribute.
 
 use darling::Error;
+use proc_macro::TokenStream;
 use proc_macro2::TokenStream as TokenStream2;
 use quote::{quote, quote_spanned};
 use syn::{meta::ParseNestedMeta, parse_quote, spanned::Spanned, Expr, Result};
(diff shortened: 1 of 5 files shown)


host-macros/src/server.rs
```diff
@@ -33,15 +31,15 @@ impl ServerArgs {
                 let buffer = meta.value().map_err(|_| Error::custom("mutex_type must be followed by `= [type]`. e.g. mutex_type = NoopRawMutex".to_string()))?;
                 self.mutex_type = Some(buffer.parse()?);
             }
-            "attribute_data_size" => {
-                let buffer = meta.value().map_err(|_| Error::custom("attribute_data_size msut be followed by `= [size]`. e.g. attribute_data_size = 32".to_string()))?;
-                self.attribute_data_size = Some(buffer.parse()?);
+            "attribute_table_size" => {
+                let buffer = meta.value().map_err(|_| Error::custom("attribute_table_size msut be followed by `= [size]`. e.g. attribute_table_size = 32".to_string()))?;
```
jamessizeland, line 35: Typo there from older code "msut"

jamessizeland (approved):

jamessizeland:
Beautiful, and very nice code comments.

host-macros/src/server.rs
petekubiak: Good eye

lulf (approved):

PR #177, opened 2024-11-25
Add HID appearance

HaoboGu:
Added some common appearance for HID, fix the wrong GAMEPAD appearance code

### DIFF
--- a/host/src/gap.rs
+++ b/host/src/gap.rs
@@ -51,8 +51,10 @@ pub mod appearance {
     ///
     /// ```rust
     /// use trouble_host::prelude::*;
-    ///
-    /// const GAMEPAD: &[u8; 2] = &appearance::new(0x00F, 0x040);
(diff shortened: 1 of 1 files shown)


HaoboGu:
Thanks for the suggestion! I've added the prefix and updated the example code

lulf:
Thanks for the PR! I think rather than adding more of these values to trouble, we should start to use the values from bt-hci https://github.com/embassy-rs/bt-hci/blob/main/src/uuid/appearance/categories.rs

HaoboGu:
> Thanks for the PR! I think rather than adding more of these values to trouble, we should start to use the values from bt-hci https://github.com/embassy-rs/bt-hci/blob/main/src/uuid/appearance/categories.rs

wow I didn't know that! Since we already have those values from bt-hci, I'm closing this PR

PR #179, opened 2024-11-25
Fix visibility constraint for ServiceBuilder's ATTRIBUTE_COUNT constant

plaes:
This makes it possible to import derived gatt_service from separate modules.


lulf (approved):
Thanks!

jamessizeland:
nice catch!

PR #181, opened 2024-11-27
use bt_hci BluetoothUuid16

jamessizeland:
I've brought in the BluetoothUuid16 NewType from bt_hci and made the constant UUIDs available through trouble_host::prelude.  I guess there is an open question on whether to unify the trouble_host::Uuid with this further or not?

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -58,7 +58,7 @@ where
         stack,
         GapConfig::Peripheral(PeripheralConfig {
             name: "TrouBLE",
-            appearance: &appearance::GENERIC_POWER,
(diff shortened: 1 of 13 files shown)


lulf:
I think for now it's sufficient to make Uuid a non-public type

jamessizeland:
I suggest we leave it public for now as privitising it breaks methods such as GattClient::services_by_uuid(), and to swap over we'd want a way of expressing BluetoothUuid128 too?

petekubiak (approved):

lulf:
> I suggest we leave it public for now as privitising it breaks methods such as GattClient::services_by_uuid(), and to swap over we'd want a way of expressing BluetoothUuid128 too?

Good point, I forgot about that api. Just leave it then.

PR #187, opened 2024-11-29
Make example panic if advertise fails

lulf:


examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -74,7 +74,11 @@ where
                     // then return to advertising state.
                     select(connection_task, counter_task).await;
                 }
-                Err(_) => info!("[adv] error"),
+                Err(e) => {
+                    #[cfg(feature = "defmt")]
+                    let e = defmt::Debug2Format(&e);
```
plaes, line 79: Would this work even without defmt?

lulf: I'm not sure what you mean, because it does compile on the non-defmt examples

plaes: Ok, so `defmt::Debug2Format` fixes the value for panic formatter.. Apologies for the noise.

plaes (approved):

Issue #175, opened 2024-11-22
Add default_value and descriptor to characteristic macro


HaoboGu:
Current gatt service macro fills 0s as the `AttributeData`:

https://github.com/embassy-rs/trouble/blob/224f37194d593cd884f736c0f168b097f9b61a5d/host-macros/src/service.rs#L149

It would be very useful if users could set the value of `AttributeData`.

I found the following code blocks in `host-macro`, it seems `value` and `descriptor` macro args are not implemented. May I ask that is there any plan to finish this?

https://github.com/embassy-rs/trouble/blob/224f37194d593cd884f736c0f168b097f9b61a5d/host-macros/src/characteristic.rs#L115-L126

jamessizeland:
It is on the to-do list yes 😊, you're welcome to PR an implementation, but also good to know it's a wanted feature so we can prioritise it.

HaoboGu:
yeah, this is commonly used in HID over BLE, for example, setting hid  info and report map

lulf:
Agreed, I think that could be useful for the device information service as well.

jamessizeland:
Do we have a preference on the interface for descriptors? I can't work out how they are expected to look in Softdevice but I was wondering if they should be also part of the #[characteristic(...)] or if we should also have #[descriptor(...)] too and allow multiple of them per characteristic.

@petekubiak  and I have got the default value stuff working with nice type checking and error messages, I'll open a draft now to show that.

lulf:
It should be possible to use multiple descriptors per characteristic (in fact someone is already doing that, see https://github.com/embassy-rs/trouble/issues/95)

jamessizeland:
cool, I'll take a look at that, I hadn't spotted thanks.

jamessizeland:
ah sorry on the multiple descriptors front I meant either something like:

```rust
// Battery service
#[gatt_service(uuid = "180f")]
struct BatteryService {
    #[characteristic(uuid = "2a19", read, write, notify, descriptior = (uuid = "1234", value = "something"), descriptor = (uuid = "2345", value = "something else"))]
    level: u8,
}
```

or

```rust
// Battery service
#[gatt_service(uuid = "180f")]
struct BatteryService {
    #[characteristic(uuid = "2a19", read, write, notify)]
    #[descriptor(uuid = "1234", value = "something")]
    #[descriptor(uuid = "2345", value = "something else")]
    level: u8,
}

```

lulf:
[...]

PR #190, opened 2024-12-09
Add descriptors and default values

jamessizeland:
Discussed in https://github.com/embassy-rs/trouble/issues/175

- [x]  default values with type checking
- [x] descriptors
- [x] docstrings


jamessizeland:
parsing towards this api:
```rust
#[gatt_service(uuid = "180f")]
struct BatteryService {
    /// Test for the sake of it
    /// One two three
    #[characteristic(uuid = "2a19", read, write, notify, value = 88, on_read = battery_level_on_read)]
    #[descriptor(uuid = "2a20", read, write, notify, value = "Hello")]
    #[descriptor(uuid = "2a21", read, write, notify, value = "Hello Again")]
    level: u8,
}
```

jamessizeland:
This is what the API looks like now, I've added loads of custom errors in for when you write the wrong thing, thanks @petekubiak for your help on some of the arcane macro errors! I'm pleased with this, let me know if there's anything you want me to explain @lulf, @HaoboGu.

One thing I'm not 100% sure on, is I've made a call on how to deal with writeable descriptors that you need to specify a capacity.  This could be done in a few other ways so I'd like feedback on if that's a good idea or not.

Have a go at trying to write the wrong thing in the struct definition and see if there's any edge cases we've missed!

```rust
const VAL: &str = "Hello World";

/// Battery service
#[gatt_service(uuid = "180f")]
struct BatteryService {
    /// Test for the sake of it
    /// One two three
    #[characteristic(uuid = "2a19", read, write, notify, value = 99, on_read = battery_level_on_read)]
    #[descriptor(uuid = "2a20", read, write, notify, capacity = 100)]
    #[descriptor(uuid = "2b20", read, notify, value = "Demo description")]
    level: u8,
    #[characteristic(uuid = "2a22", read, write, notify, value = 42.0)]
    #[descriptor(uuid = "2a21", read, write, notify, value = VAL, capacity = VAL.len() as u8)]
    rate_of_discharge: f32,
}
```

host-macros/src/lib.rs
```diff
@@ -144,25 +151,104 @@ pub fn gatt_service(args: TokenStream, item: TokenStream) -> TokenStream {
...
+    let mut characteristic_count = 0;
+    for attr in &field.attrs {
+        if let Some(ident) = attr.path().get_ident() {
+            match ident.to_string().as_str() {
+                "doc" => {
+                    if let Ok(meta_name_value) = attr.meta.require_name_value() {
+                        if let syn::Expr::Lit(value) = &meta_name_value.value {
+                            if let Some(text) = &value.lit.span().source_text() {
+                                let text: Vec<&str> = text.split("///").collect();
+                                if let Some(text) = text.get(1) {
```
petekubiak, line 198: So you're starting with a doc string line, you split at `///` to get anything before the '///' as text[0] and anything after as text[1]?
Would this filter out "blank" docstring lines, as in 
```rust
///
```
and if so is that what we want to do?

host-macros/src/lib.rs
```diff
@@ -144,25 +151,104 @@ pub fn gatt_service(args: TokenStream, item: TokenStream) -> TokenStream {
...
+                    Ok(args) => descriptors.push(args),
+                    Err(e) => {
+                        *err = Some(e);
+                        return REMOVE; // If there was an error parsing the descriptor, remove the field.
+                    }
+                },
+                "characteristic" => {
+                    // make sure we only have one characteristic meta tag
+                    characteristic_count += 1;
+                    if characteristic_count > 1 {
```
petekubiak, line 218: Could this be a `bool` instead?

host-macros/src/service.rs
```diff
@@ -129,10 +130,60 @@ impl ServiceBuilder {
...
+            .map(|(index, args)| {
+                let name_screaming =
+                    format_ident!("DESC_{}_{index}", to_screaming_snake_case(characteristic.name.as_str()));
+                let properties = set_desc_access_properties(&args);
+                let uuid = args.uuid;
+                // let read_callback = args.on_read.as_ref();
+                // let write_callback = args.on_write.as_ref();
+                let writeable = args.write || args.write_without_response;
+                let capacity = match &args.capacity {
+                    Some(cap) => quote!(#cap),
```
petekubiak, line 147: TIL github diff is confused by #

host-macros/src/service.rs
```diff
@@ -129,10 +130,60 @@ impl ServiceBuilder {
...
+                    None => {
+                        if writeable {
+                            panic!("'capacity' must be specified for writeable descriptors");
+                        }
+                        quote!(0u8)
```
petekubiak, line 152: Shouldn't the capacity be set to the length of the default value (if available) or some predefined size if it's set dynamically? (e.g. 32 bytes or something)

lulf:
Nice work!

I'm not sure what to think about the 'calling functions to initialize' thing. Maybe  it's a clearer API if there is an API on the characteristic allowing you to set the value of a descriptor.

host-macros/src/service.rs
```diff
@@ -147,17 +198,24 @@ impl ServiceBuilder {
             .on_write
             .as_ref()
             .map(|callback| quote!(builder.set_write_callback(#callback);));
+        let default_value = match characteristic.args.default_value {
+            Some(val) => quote!(#val),      // if set by user
+            None => quote!(#ty::default()), // or default otherwise
```
petekubiak, line 203: I wonder if we should constrain `GattValue` and `FixedGattValue` to be `Default`? Otherwise one might have some interesting errors if they used a type which doesn't implement `Default`

host-macros/src/service.rs
```diff
@@ -198,23 +256,34 @@ impl ServiceBuilder {
...
-
         // Processing common to all fields
-        for field in fields {
+        for (field, doc_string) in fields.iter().zip(doc_strings) {
+            let docs: TokenStream2 = doc_string
+                .lines()
+                .map(|line| {
+                    let span = field.span();
+                    quote_spanned!(span=>
+                        #[doc = #line]
```
petekubiak, line 278: Oh I see, you're removing the syntactic sugar - is that why the docstrings weren't showing up in the tooltips?

host-macros/src/service.rs
```diff
@@ -242,3 +311,18 @@ fn set_access_properties(args: &CharacteristicArgs) -> Vec<TokenStream2> {
     parse_property_into_list(args.indicate, quote! {CharacteristicProp::Indicate}, &mut properties);
     properties
 }
+
+/// Parse the properties of a characteristic and return a list of properties
```
petekubiak, line 315: Should this be "descriptor" rather than "characteristic"?

host/tests/gatt_derive.rs
```diff
@@ -25,7 +25,7 @@ struct Server {
 
 #[gatt_service(uuid = "408813df-5dd4-1f87-ec11-cdb000100000")]
 struct CustomService {
-    #[characteristic(uuid = "408813df-5dd4-1f87-ec11-cdb001100000", read, write, notify, on_read = value_on_read, on_write = value_on_write)]
+    #[characteristic(uuid = "408813df-5dd4-1f87-ec11-cdb001100000", read, write, notify, on_read = value_on_read, on_write = value_on_write, value = 42)]
```
petekubiak, line 28: Maybe add some descriptors and docstrings in here for good testing measure?

petekubiak (changes requested):
Really nice work - it's looking great!

host-macros/src/lib.rs
jamessizeland: Good question, we do not want to filter them out.

host-macros/src/service.rs
jamessizeland: That assumes the default value is the longest it'll ever be though?

host-macros/src/service.rs
jamessizeland: I'm not sure, they still show up as sugared in the expanded output.

host-macros/src/service.rs
jamessizeland: Looks like we can't do this, because of impl GattValue for [u8; N]

So might need to change back what we do in the None case.

host-macros/src/lib.rs
jamessizeland: Ok I've tried it and it does still allow empty lines with this code. It would just deny an 'initial' empty line.

host-macros/src/service.rs
petekubiak: Sorry what I mean is that if the capacity isn't explicitly set by the user we need to make some sort of call on what the capacity should be.
My understanding is that the most common use case for descriptors is that they don't change for the lifetime of the application. If this is accurate, I think the most ergonomic and memory efficient way to handle this is to size the descriptor's memory to the given value. An alternative would be to just have a default capacity, however if we do that it will mean that in almost all cases we are either wasting memory, or causing the user to have to realise their default value is being truncated, and then figure out how to change the capacity.
Either way though, it looks like the code here is setting the capacity to 0, so nothing will be stored?

host-macros/src/service.rs
petekubiak: Yeah I think we'll have to revert it, otherwise [u8; N] is unusable.

host-macros/src/service.rs
petekubiak: Huh.

host-macros/src/service.rs
jamessizeland: solved! `<#ty>::default()`

host-macros/src/service.rs
jamessizeland: you are correct, I've fixed this now by:

```rust
let default_value = match &args.default_value {
                    Some(val) => quote!(#val), // if set by user
                    None => quote!(""),
                };
                let capacity = match &args.capacity {
                    Some(cap) => quote!(#cap),
                    None => {
                        if writeable {
                            panic!("'capacity' must be specified for writeable descriptors");
                        }
                        quote!(#default_value.len() as u8)
                    },
                };
```

petekubiak (changes requested):

host-macros/src/service.rs
```diff
@@ -137,32 +149,35 @@ impl ServiceBuilder {
...
-                };
-                // At least two attributes will be added to the attribute table for each characteristic:
-                // - The characteristic declaration
-                // - The characteristic's value declaration
-                //
-                // If the characteristic has either the notify or indicate property,
-                // a Client Characteristic Configuration Descriptor (CCCD) declaration will also be added.
-                self.attribute_count += if args.notify { 3 } else { 2 };
+
+                self.increment_attributes(access);
```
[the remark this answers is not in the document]
jamessizeland: @lulf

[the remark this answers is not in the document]
alexmoon: >     1. if descriptors can be `notify` and/or `indicate`, do they also need a CCCD attribute to be pushed on to the table?

No, descriptors can't have descriptors (such as CCCD) and can't be notify or indicate.

>     2. currently this will also increment the attribute count by 2 / 3 for descriptors. Should we be only incrementing by 1 (or 2 with CCCD) for descriptors instead?

Yes, descriptors should only have a single attribute.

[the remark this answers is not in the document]
jamessizeland: Nice, I'll make that change, thanks!

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -24,7 +24,10 @@ struct Server {
 // Battery service
 #[gatt_service(uuid = "180f")]
 struct BatteryService {
-    #[characteristic(uuid = "2a19", read, write, notify, on_read = battery_level_on_read, on_write = battery_level_on_write)]
+    /// Battery Level
+    #[descriptor(uuid = "2b20", read, write, value = "Battery Level", capacity = 100, on_read = battery_level_on_read)]
```
lulf, line 28: This should probably not be `write`?

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -24,7 +24,10 @@ struct Server {
...
+    #[descriptor(uuid = "2b21", read, write, value = "Other Descriptor", capacity = 255, on_write = battery_level_on_write)]
+    #[characteristic(uuid = "2a19", read, write, notify, on_read = battery_level_on_read, on_write = battery_level_on_write, value = 10)]
```
lulf, line 30: I'm thinking it's relying too much on macros here, and that maybe this is a signal it should be designed differently. What if one could define descriptors like this:

```
#[characteristic(uuid = "2a19", read, write, notify, on_read = battery_level_on_read, on_write = battery_level_on_write, value = 10)]
level: u8,

#[descriptor(uuid = "2b21", read, write, value = "Other Descriptor")
level_descriptor: Vec<u8, 255>,

```

What's missing in this example is how to tie the descriptor to the characteristic. Maybe one could represent it like this:

```
#[characteristic(uuid = "2a19)]
level: Characteristic {
    #[value(read, write, notify, on_read = battery_level_on_read, on_write = battery_level_on_write, value = 10)]
    value: u8,
    #[descriptor(uuid = "2b21", read, write, value = "Other Descriptor")
    descriptor: Vec<u8, 255>,
}
```

jamessizeland: Hmm, I don't know about these alternatives, I feel they add more complexity to the simple case (no descriptors) or decouple the characteristic and descriptor.

We could reduce some visual noise by shortening the names of the inputs? i.e. 'read' becomes 'r' etc.

@petekubiak made a good point though that we don't currently have a unique name per descriptor for the get and set methods, so that's still a challenge to make ergonomic.

lulf: I think the simple case could still be as is today. This more complex form can be used only if complex descriptors are needed.

Shortening the name of the inputs would make the code harder to read, so would be against that.

With the above, it should be simple to generate appropriate names for setting and getting the descriptor values, since the macro can use the descriptor field name.

petekubiak (approved):

examples/apps/src/ble_bas_peripheral.rs
petekubiak: At first I thought that the struct variant would require a whole new proc-macro to deal with the characteristic structure, but it's just occurred to me that we can parse the input token stream however we want, so we can use the existing `#[gatt_service]` macro to parse the Characteristic `structure` and flatten the generated fields into the service struct.
I think this is a good solution in that it allows us to both associate the descriptors with their characteristics and ergonomically provide a name for the descriptor for use with `get()` and `set()`.

jamessizeland:
Do you think it's worth me tidying this up and refocusing it to just read only descriptors for now, so we can close it off? Then come back to improving the API in a future PR?

lulf:
Sounds good to me!

petekubiak:
> Do you think it's worth me tidying this up and refocusing it to just read only descriptors for now, so we can close it off? Then come back to improving the API in a future PR?

Yep, sounds good!

jamessizeland:
ok I'm happy to call this one done and ready to add smaller PRs onto.

jamessizeland:
Apologies for the spam on this! The test was broken, so I'll get it back to a working state.

PR #201, opened 2024-12-17
Update esp32 example for non esp32c3 targets

bytedream:
This PR changes several things in the the esp32 example:
- Set `espflash flash --monitor` to be the runner for all esp related targets
- Add targets as comments for different chips in config file
- Add `esp32` feature for normal ESP32 chip
- Add cargo alias to run example on every supported esp32 chip
- Decrease `embassy-executor` task arena size to `task-arena-size-32768`, any higher value will cause an compile error on ESP32
- Remove `esp32s2` feature, this chip doesn't support Bluetooth at all
- Remove hardcoded `esp32c3` feature in `esp-backtrace` dependency


jamessizeland:
might be worth adding some alias lines in .cargo/config.toml, we tend to do this for multi-chip projects, i.e.

[alias]
esp32c3 = "run --release --features=esp32c3 --target=riscv32imc-unknown-none-elf"
esp32c6 = "run --release --no-default-features --features=esp32c6 --target=riscv32imac-unknown-none-elf"

bytedream:
Thanks for the suggestions I'll add it. I didn't knew that an cargo alias could be specified in the config, that's neat

lulf (approved):
Excellent, thanks!

PR #202, opened 2024-12-18
Add more esp32 examples

bytedream:
This PR adds more examples for the esp32 platform.

It introduces an optional `esp32` feature to the `examples/apps` crate that influences the hardcoded L2CAP MTU and is intended to be set by the esp32 examples only. The current MTU of `128` causes a `BleHost(Hci(Unsupported Feature or Parameter Value))` error here when using an esp: https://github.com/embassy-rs/trouble/blob/aaf2d02b09bc4ecdc4cd3a820cf9e268500b5082/host/src/host.rs#L813-L820
This is a known limitation, as seen in an issue comment over at the esp-idf repo https://github.com/espressif/esp-idf/issues/942#issuecomment-336809446. They wrote that any total host buffer size <= 1021 bytes aren't safe to use (1017 bytes ACL MTU + 4 bytes ACL Data Packet header) as long as ACL packet segmentation isn't implemented, so I've set the MTU to 1017 if the `esp` feature is activated.
I tested my changes with an ESP32C3 and ESP32S3 and in both cases I was able to use a lower MTU than 1017 but it still failed at 128 (200 failed too, 512 not)

### DIFF
--- a/examples/apps/Cargo.toml
+++ b/examples/apps/Cargo.toml
@@ -28,6 +28,7 @@ defmt = [
     "embedded-io/defmt-03",
     "embedded-hal/defmt-03"
 ]
+esp = []
 log = [
     "dep:log",
     "trouble-host/log",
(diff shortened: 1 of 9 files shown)


lulf (approved):
Thanks!

Issue #215, opened 2025-01-03
Cannot Use Heapless Vec with Capacity > 32 Due to Default Impls


pperanich:
With the current implementation of variable-length fields (when using heapless `Vec`s), you are limited to a `Vec` of length 32 or less. This is because the underlying storage is a `[u8; N]`, and `Default::default()` is used to initialize the static cell. However, the `Default` trait in the standard library is [only implemented for `[u8; 32]` and smaller](https://doc.rust-lang.org/nightly/core/default/trait.Default.html#impl-Default-for-%5BT;+32%5D). Attempting to use a larger heapless `Vec` leads to a compile error (rust-lang/rust#61415).

jamessizeland:
Thanks for spotting this! Do you intend to PR a fix or would you like us to look at it?

PR #216, opened 2025-01-03
fix: initialize storage arrays manually instead of `Default::default()`

pperanich:
Fixes embassy-rs/trouble#215


host/tests/gatt_derive.rs
```diff
@@ -40,6 +40,8 @@ struct CustomService {
     pub third: [u8; 2],
     #[characteristic(uuid = "408816df-5dd4-1f87-ec11-cdb001100000", read, write, notify)]
     pub fourth: heapless::Vec<u8, 3>,
+    #[characteristic(uuid = "408816df-5dd4-1f87-ec11-cdb001100000", read, write, notify)]
```
jamessizeland, line 43: We're not currently checking for uniqueness of UUIDs within a service (tbd) but can you make this unique all the same please?

pperanich: Good call. Would probably be a good idea to add a check sometime in the future.

lulf (approved):

PR #219, opened 2025-01-04
add rssi task to gatt periph example

jamessizeland:
We might not want to merge this, but it could be useful as an example.

Tested only on esp32c3 as that's all I've got to hand right now.


lulf (commented):
How about moving that into the counter_task, just printing the RSSI for every tick? We should probably add the ControllerCmdSync<ReadRSSI> to https://github.com/embassy-rs/trouble/blob/main/host/src/lib.rs#L234 perhaps, I think all controllers should support that.

lulf (approved):
LGTM, thanks!

PR #213, opened 2025-01-02
Update bt-hci, nrf-sdc, embassy, esp-hal.

Dirbaio:
Draft because it depends on

- esp-hal to update to embassy-time v0.4 https://github.com/esp-rs/esp-hal/pull/2701
- ~~https://github.com/bugadani/esp-hal/pull/4~~
- ~~https://github.com/alexmoon/nrf-sdc/pull/42~~
- ~~https://github.com/alexmoon/nrf-sdc/pull/41~~

PR #223, opened 2025-01-08
ESP32: Last three to drop?

lure23:
Some pump-ups of dependencies (not `bc-hci` and `embassy-executor`, though).

More importantly, asking whether the `[patch.crates-io]` is needed any more (after `esp-hal` 0.22 release)?


lure23:
Might be related to: https://github.com/embassy-rs/trouble/pull/213

lure23:
After #213 is merged, there's little remaining here. Three dependencies that seem unneeded to me.

Would someone comment whether those are actually needed. If not, I can make a separate short PR to just wipe them out.

lulf (approved):
LGTM, thanks!

lure23:
Many of the changes originally suggested have been applied via other PR's. The issue of the three remaining dependcies remains. I merely point out that the examples compile and run (on ESP32-C6) without them, so wondering whether they can also be slimmed down.
 
This would be the end of my ESP32 diet! 😎

lure23:
One question on this, @lulf if you may..

```
$ git pull
...
Fast-forward
 examples/apps/Cargo.toml       | 1 -
 examples/esp32/Cargo.lock      | 4 ----
 examples/esp32/Cargo.toml      | 2 --
 examples/serial-hci/Cargo.lock | 1 -
 4 files changed, 8 deletions(-)
```

What's the policy with the Cargo.lock files? Why was `serial-hci/Cargo.lock` affected by me making the PR?

I understand the purpose of lock files (from Node.js); just need to understand when to commit them; when not. Thank you

jamessizeland:
Rust-analyzer does more background work than nodejs does, so minor or bugfix semver updates can occur like this without something like an `npm update`. I'm fine with this, if the tests pass, these dependency updates should be pulled in at some point anyway.

lulf:
Libraries (like trouble-host) do not have lock files. Binaries (examples) have lock files. This ensures that external binaries can have more flexibility, while examples have a high chance of working.

If you ran ci.sh, it will update lock files (you need to pass --frozen for it not to)

PR #226, opened 2025-01-09
Esp32 comments //not for merge

lure23:
This PR is intended to be a contextual conversation on some parts of the `esp32` samples. I collected questions / opinions while going through the code.


examples/apps/Cargo.toml
```diff
@@ -12,7 +12,6 @@ embassy-futures = "0.1.1"
 embassy-sync = { version = "0.6" }
 embassy-time = "0.4"
 embedded-hal = "1.0"
-embedded-hal-async = "1.0"
```
lure23, line 15: Oops. Also mentioned in #223 - is it needed?

examples/esp32/Cargo.toml
```diff
@@ -14,8 +14,8 @@ esp-println = { version = "0.12.0", features = ["log"] }
 esp-wifi = { version = "0.11.0", features = [ "ble" ] }
 trouble-example-apps = { version = "0.1.0", path = "../apps", features = ["esp", "log"] }
 bt-hci = { version = "0.2" }
-embassy-futures = "0.1.1"
-embassy-time = { version = "0.4", features = ["generic-queue-8"] }
+#embassy-futures = "0.1.1"
+#embassy-time = { version = "0.4", features = ["generic-queue-8"] }
```
lure23, line 18: Duplicated in #223

lure23 (commented):
A start for discussions

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -38,9 +38,13 @@ pub async fn run<C>(controller: C)
...
-    // Using a fixed seed means the "random" address will be the same every time the program runs,
-    // which can be useful for testing. If truly random addresses are required, a different,
-    // dynamically generated seed should be used.
+    // Using a fixed "random" address can be useful for testing. In real scenarios, one would
+    // use e.g. the MAC 6-byte array as the address (how to get that varies by the platform).
+    //
+    // tbd. How about providing the address from the caller (i.e hardware specific code)? That way,
+    //      the code to provide a chip-specific MAC could be shown. Furthermore, could the selection
+    //      of fake-vs-real address be a feature ('fixed-address-testing').
+    //
```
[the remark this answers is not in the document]
lulf: Yeah, seed is a bad term, I agree it should be removed. Passing it in as a parameter to the app is certainly an option, and reading them from the hardware mac.

The reason it's hardcoded is this: you can spin up ble_bas_peripheral and ble_bas_central on two random types of hardware, and they will talk to each other. If we use the hardware mac, it's not as neat 'just works' experience. 

The other advantage of fixed, is that the recently added HIL tests can now work with the example without modifications (otherwise it would have to be overriden during the CI build or something to use a fixed address).

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -55,6 +59,12 @@ where
         appearance: &appearance::power_device::GENERIC_POWER_DEVICE,
     }))
     .unwrap();
+
+    // Below could use 'join' in the same way as 'ble_bas_central' and 'ble_l2cap_peripheral' do?
+    // i.e. 'ble_task(runner)' comes first, the 'app_task' doesn't need to be named....
+    // Are 'join' and 'select' interchangeable, in this way???
+    // Benefit: 'join' use looks cooler, clearer.
+
```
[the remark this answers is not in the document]
lulf: You're right, in this case it doesn't matter if it's a select or join, because these two tasks run forever (or until panic).

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -102,6 +113,9 @@ async fn ble_task<C: Controller>(mut runner: Runner<'_, C>) {
     }
 }
 
+// Consistency with the 'panic!' and 'info!' tags, perhaps?
+// 'conn_task' or 'conn' instead of 'gatt'? Likely remains from previous task naming? Compare to 'ble_task', above.
+
```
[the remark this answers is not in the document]
lulf: Yeah, the tags are just for being able to associate the task. The loop was moved from an earlier task named 'gatt', hence the name. So I think it's fine to change it :+1:

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -156,7 +169,7 @@ async fn advertise<'a, C: Controller>(
     AdStructure::encode_slice(
         &[
             AdStructure::Flags(LE_GENERAL_DISCOVERABLE | BR_EDR_NOT_SUPPORTED),
-            AdStructure::ServiceUuids16(&[Uuid::Uuid16([0x0f, 0x18])]),
+            AdStructure::ServiceUuids16(&[Uuid::Uuid16([0x0f, 0x18])]),     // Q: what is this???  Connected to us being 'bas'; what would I place there for a custom service?
```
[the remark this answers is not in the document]
lulf: You would usually put some the uuid of your main service there, to allow scanners to identify your device before connecting to it. But it's not required.

examples/apps/src/ble_l2cap_peripheral.rs
```diff
@@ -86,3 +86,15 @@ where
...
+
+// General Q:
+//  What is the purpose of L2CAP in the examples?  I get that there is a 'bas' example, but what
+//  would be the case where a user of 'TrouBLE' wants to deal with the (did I get it right?) lower
+//  level of abstraction?
+//
+//  What may make this problematic is the layout having both 'ble' and 'l2cap' alongside each other,
+//  as siblings. When one likely is a higher abstraction; 'bas' using 'l2cap' underneath (just
+//  assuming; please correct me or point to a learning resource?).
+
```
lulf, line 98: In BLE, you can use l2cap connection oriented channels (as done in the examples), completely without gatt. GATT is a protocol that is built on top of a single l2cap channel, so it's a different level of abstraction, which includes all of the standardized GATT characteristics etc.

Using l2cap connection oriented channels is nice if you just want to send messages between a peripheral and central. It might be useful to think of it as a UDP (though with more reliability guarantees and flow control).

examples/esp32/.cargo/config.toml
```diff
@@ -16,14 +16,31 @@ runner = "espflash flash --monitor"
...
 # target = "riscv32imac-unknown-none-elf" # ESP32-C6 and ESP32-H2
 
+  # ^-- I find the above section a bit.. difficult to understand (without documentation) and would propose removing
+  #   it. There seem to be two mechanisms for selecting the target: the alias ('cargo esp32{..}') and this '[build]'
+  #   section that requires modifying the file. I'd like to see target selection not need editing.
+  #
+  #   Personally, I use:
+  #     <<
+  #       $ cargo build --release --no-default-features --features=esp32c6 --target=riscv32imac-unknown-none-elf --bin ble_l2cap_peripheral
+  #     <<
```
lulf, line 26: Yeah, I'm open to remove that.

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -92,6 +102,7 @@ where
 ///
 /// spawner.must_spawn(ble_task(runner));
 /// ```
+// Uncertain, what the '.must_spawn()' does; is this a common Embassy pattern?
```
lulf, line 105: Must spawn will panic if there is no free 'task handles' in the task pool. It's basically the same as spawn().unwrap().

lulf (commented):
Thanks for the questions! I've tried to answer as best as I can (and apologies if I forgot some).

jamessizeland:
Just read through, some good questions. I appreciate the effort to reduce the introduction barriers here!

PR #228, opened 2025-01-09
small changes proposed to 'esp32' build and 'ble_bas_peripheral' example

lure23:
Some things discussed in #226 collected as a mergeable PR.

- use of `join` in `apps/ble_bas_peripheral` example
- change to `Address` comment; "seed" gone
- more consistency to logging tags
- `[build]: target` removed from `.cargo/config.toml`
- `[unstable]` -""-

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -1,4 +1,4 @@
-use embassy_futures::select::select;
+use embassy_futures::{join::join, select::select};
 use embassy_time::Timer;
 use trouble_host::prelude::*;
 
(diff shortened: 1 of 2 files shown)


jamessizeland (approved):
Nice tidying up!

jamessizeland:
I think the seed comment was from me as I was thinking of a random generator that's set with a fixed seed giving the same "random" result every time. But yeah but clearer without it.

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -111,7 +111,7 @@ async fn conn_task(server: &Server<'_>, conn: &Connection<'_>) -> Result<(), Err
     loop {
         match conn.next().await {
             ConnectionEvent::Disconnected { reason } => {
-                info!("[gatt] disconnected: {:?}", reason);
```
jamessizeland, line 114: I'd prefer these stay as [gatt] or something generic, as I like to copy this whole function into other projects and know these console logs are referring to the ble service and not something else.

But not a particularly strong opinion.

lure23: Sure. My point in #226 was merely that as they now are, they don't seem consistent to a newcomer.

Should I revert to not changing tags, in this PR?  Or is it just the ones within `conn_task` -> `gatt` (some of them were different)?

jamessizeland: Yeah I think that works better for me. Like I say it's not a strongly held opinion though.

lulf (approved):
LGTM, thanks!

PR #230, opened 2025-01-11
Refactor AttErrorCode as wrapper type around u8

petekubiak:
This change allows the entire range of u8 as error codes. According to the Bluetooth spec the entire range 0 - 255 is valid, however only a subset of these are defined in the Bluetooth spec.
Some are reserved for application error codes, some are reserved for common profile and service error codes, and some are reserved for future use.

Still to do:
- [ ] Match arm for application error codes
- [ ] Match arm for common error codes
- [ ] Update "other" match arm

### DIFF
--- a/host/src/att.rs
+++ b/host/src/att.rs
@@ -1,5 +1,5 @@
-use num_enum::TryFromPrimitive;
-use thiserror::Error;
+use core::fmt::Display;
+use core::mem;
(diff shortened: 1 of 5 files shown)


lulf (approved):

jamessizeland (approved):
Nice work

PR #235, opened 2025-01-12
Increase mtu for ble_bas_peripheral for esp examples

### LINKED ISSUE #229: Panic on Peripheral::advertise when using esp32c6
When the reaching here:

https://github.com/embassy-rs/trouble/blob/419e29886ee696570f0513b76b7aa52f82a19ff5/host/src/peripheral.rs#L64-L73

the BleHost:command is called, but if fails on OnceLock on initalized

```rust
    pub(crate) async fn command<C>(&self, cmd: C) -> Result<C::Return, BleHostError<T::Error>>
    where
        C: SyncCmd,
        T: ControllerCmdSync<C>,
    {
        let _ = self.initialized.get().await;

        let ret = cmd.exec(&self.controller).await?;
        Ok(ret)
    }
```

And the panic:

```
INFO - Our address = Address { kind: AddrKind(1), addr: BdAddr([65, 90, 227, 30, 131, 231]) }
INFO - Starting advertising and GATT service
INFO - [host] filter accept list size: 12
INFO - [host] setting txq to 24, fragmenting at 255
INFO - [host] configuring host buffers (8 packets of size 251)


panicked at /tmp/trouble_bk/examples/apps/src/ble_bas_peripheral.rs:104:13:
[ble_task] error: BleHost(Hci(Invalid HCI Command Parameters))

Backtrace:

0x4200e91e
0x4200e91e - embassy_executor::raw::util::SyncUnsafeCell<T>::get
    at /home/eloy/.cargo/registry/src/index.crates.io-6f17d22bba15001f/embassy-executor-0.7.0/src/raw/util.rs:55
0x42016b90
0x42016b90 - main
    at /tmp/trouble_bk/examples/esp32/src/bin/ble_bas_peripheral.rs:11
0x420055e8
0x420055e8 - hal_main
    at /home/eloy/.cargo/git/checkouts/esp-hal-42ec44e8c6943228/7288d03/esp-hal/src/lib.rs:485
```

I might fix it, but I need to check how to use gdb with espflash, etc.. it might take a while, so any hint it's welcome :-)

Using bas_peripheral_example and esp32c6

sjroe:
Fixes #229

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -2,8 +2,12 @@ use embassy_futures::{join::join, select::select};
 use embassy_time::Timer;
 use trouble_host::prelude::*;
 
-/// Size of L2CAP packets (ATT MTU is this - 4)
-const L2CAP_MTU: usize = 251;
+/// Size of L2CAP packets
+#[cfg(not(feature = "esp"))]
+pub const L2CAP_MTU: usize = 128;
+#[cfg(feature = "esp")]
+// Some esp chips only accept an MTU >= 1017
+pub const L2CAP_MTU: usize = 1017;
 
 /// Max number of connections
 const CONNECTIONS_MAX: usize = 1;
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -4,7 +4,7 @@ use trouble_host::prelude::*;
 
 /// Size of L2CAP packets
 #[cfg(not(feature = "esp"))]
-pub const L2CAP_MTU: usize = 128;
+pub const L2CAP_MTU: usize = 251;
 #[cfg(feature = "esp")]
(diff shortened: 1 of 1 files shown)


examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -2,8 +2,12 @@ use embassy_futures::{join::join, select::select};
 use embassy_time::Timer;
 use trouble_host::prelude::*;
 
-/// Size of L2CAP packets (ATT MTU is this - 4)
-const L2CAP_MTU: usize = 251;
+/// Size of L2CAP packets
+#[cfg(not(feature = "esp"))]
+pub const L2CAP_MTU: usize = 128;
```
lure23, line 7: Please note that this changes the behaviour for non-`esp` compilations.  Here, the value used is `251`, unlike in the place the code was copied from.

sjroe: Ooops. Reverted to 251 for non esp.

Issue #155, opened 2024-11-04
Expand get and set methods to allow getting and setting all attributes


petekubiak:
This issue stems from [this comment](https://github.com/embassy-rs/trouble/pull/151#discussion_r1826037258) in #151.
Currently the `get` and `set` methods on the `gatt_server` macro-generated type assume that they would only be called on characteristics and not other types of attribute. This is because they are based on the methods of the same names on the `AttributeTable`. In order to allow getting and setting of other types of attribute types, these methods need to be changed to accept some sort of `AttributeHandle` type.

[Here](https://github.com/embassy-rs/trouble/pull/151#issuecomment-2444721033) is a suggestion of what this might look like.

lulf:
I've got nothing against the ability to read/write to handles that are not characteristics, but out of curiosity: what is the use case for using get and set on non-characteristic handles?  (I wasn't aware that you could even have attribute handles that were not a characteristic)

petekubiak:
To be honest I wasn't aware of this either, but in the linked comment @alexmoon said "you also need to be able to get and set descriptor values and the like"

alexmoon:
Attributes include:
- Primary Service Declaration
- Secondary Service Declaration
- Included Service Declaration
- Characteristic Declaration
- Characteristic Extended Properties
- Characteristic Value
- Descriptor Value

I believe from the client's perspective all of those are read-only except for Characteristic Value and Descriptor Value. Descriptor Value includes things like the Client Characteristic Configuration Descriptor (for enabling notifications/indications) but can also include arbitrary additional descriptors depending on the service (for example HID defines several descriptor values). So at the very least, the attribute table needs to include support for Descriptor Values.

The other GATT attributes may need to be changed at runtime by the server in certain circumstances. For example, a device may need to enable or disable a service depending on some configuration. And applications need to be able to send Service Changed indications to clients to indicate that things have changed. We probably don't need to go as far as putting everything in the attribute table, but we should keep in mind that the data is not necessarily static.

lulf:
Ok, not requiring static values is a good reason I think. I think the only thing you can't modify after the fact today is the service declarations.

PR #239, opened 2025-01-14
Consolidate UUID from slice functionality into TryFrom implementation

petekubiak:
While working on making get and set generic for all attributes (#155) I noticed there were multiple "from slice" implementations for the Uuid type, which would panic if the slice length didn't match what was expected.

I've consolidated these with a single TryFrom implementation instead.

### DIFF
--- a/host/src/attribute.rs
+++ b/host/src/attribute.rs
@@ -253,7 +253,7 @@ impl AttributeData<'_> {
         Ok(Self::Declaration {
             props: CharacteristicProps(r.read()?),
             handle: r.read()?,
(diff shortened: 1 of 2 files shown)


lulf (approved):
Thanks!

alexmoon:
FYI, there are also (rarely used) 32-bit Bluetooth UUIDs.

jamessizeland (approved):
Too late but looks great!

jamessizeland:
> FYI, there are also (rarely used) 32-bit Bluetooth UUIDs.

Worth adding an issue for?

PR #241, opened 2025-01-15
refactor: change api and support varying scan buffer sizes

lulf:
* Add Scanner type which wraps the central to provide scan features
* Make scanner buffer configurable size in order to handle large number of reports.
* Only add valid advertising reports to scan buffer.


lulf:
Needs embassy-sync 0.6.2

PR #254, opened 2025-01-20
Add paths to types in server and service macros

### LINKED ISSUE #245: Proc-macros do not work unless I explicitly import some stuff (as in the trouble prelude)
Consider:

```rust
// GATT Server definition
#[gatt_server]
struct Server {
    battery_service: BatteryService,
}

/// Battery service
#[gatt_service(uuid = service::BATTERY)]
struct BatteryService {
    /// Battery Level
    #[characteristic(uuid = characteristic::BATTERY_LEVEL, read, notify, value = 10)]
    level: u8,
    #[characteristic(uuid = "408813df-5dd4-1f87-ec11-cdb001100000", write, read, notify)]
    status: bool,
}
```

Once I import the `gatt_server`, `gatt_service`, `characteristic` and so on, rather than the code compiling successfully, it starts to complain that it does not know about:
`AttributeServer`
`AttributeHandle`
`GAP_SERVICE_ATTRIBUTE_COUNT`

... and so on.

I get it, that glob-importing the prelude (`use trouble_host::prelude::*;`) would fix that, but is this idiomatic Rust?
I always thought (a) preludes are optional and (b) code generated with proc-macros should always be using full type names (or do local imports) anyway - if for nothing else - due to potential collisions with the surrounding context where the macros are placed.

petekubiak:
`gatt_server` and `gatt_service` proc macros can now be used without having to add `use trouble_host::prelude::*`

Closes #245 .

### DIFF
--- a/host-macros/src/characteristic.rs
+++ b/host-macros/src/characteristic.rs
@@ -106,7 +106,7 @@ pub fn parse_uuid(meta: &ParseNestedMeta<'_>) -> Result<TokenStream> {
         let span = expr.span(); // span will highlight if the value does not impl Into<Uuid>
         Ok(quote::quote_spanned! { span =>
             {
-                let uuid: Uuid = #expr.into();
+                let uuid: trouble_host::types::uuid::Uuid = #expr.into();
                 uuid
             }
         })
--- a/host-macros/src/server.rs
+++ b/host-macros/src/server.rs
@@ -101,19 +101,19 @@ impl ServerBuilder {
         let attribute_table_size = if let Some(value) = self.arguments.attribute_table_size {
             value
         } else {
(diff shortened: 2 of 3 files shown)


lulf (approved):
LGTM! I think another improvement would be to use '::trouble_host::...` (with the `::` prefix), which ensures that it will not be confused if someone adds a mod trouble_host{} in their project.

But I think that's highly unlikely :)

ivmarkov:
Keep in mind that the current approach does not support crate renaming.

ivmarkov:
We use this guy in `rs-matter`: https://docs.rs/proc-macro-crate/latest/proc_macro_crate/

Issue #252, opened 2025-01-18
Remove `C: Controller` from `HostResources<C>`


ivmarkov:
(Mentioned ~ 3 months ago.)

I guess the idea of `HostResources` is similar to `embassy_net::StackResources` which I think is a solid approach.

However - and because `HostResources` is generified by `C` - it simply cannot be stored in a static context, or in fact in any context which has a longer lifetime than the controller itself (as the controller ~~most often than not~~ sometimes is itself not `: 'static`, or if it can be made `: 'static` via a series of `make_static` calls to its parameters it means the BT stack runs forever).

I guess it is [no coincidence](https://github.com/embassy-rs/trouble/blob/c5c2eaba230079662bc9be92e41b279492e78f58/examples/apps/src/ble_bas_peripheral.rs#L41) that the examples have the `HostResources` allocated on-stack and not from a static cell - the later is not possible IMO.

Removing the controller from the host resources will have multiple benefits:
- Resources can then be allocated in a static or in general - in a context with a longer lifetime that the controller and then re-used (pooled) for multiple controller instantiations (that's what I need)
- Resources' constant will not take `rodata` space given that the `QoS` parameter is also removed. Currently it spoils the party, as it is neither all-0-bits, nor `MaybeUninit` and is the only such one in the struct

jamessizeland:
My example project here allocates it in a static cell, which works okay where I can select the controller non-generically. https://github.com/jamessizeland/microbit-ble-gamepad/blob/main/src%2Fble%2Fgatt.rs#L49-L51

ivmarkov:
> My example project here allocates it in a static cell, which works okay where I can select the controller non-generically. https://github.com/jamessizeland/microbit-ble-gamepad/blob/main/src%2Fble%2Fgatt.rs#L49-L51

Sure but the SoftDevice controller [_is_ `'static`](https://github.com/jamessizeland/microbit-ble-gamepad/blob/abe1afc8d070c6f1b0e0d6f4fd26999c7e6c6764/src/ble/mod.rs#L20) (b) your example runs forever.

[Look at the esp32 controller](https://github.com/embassy-rs/trouble/blob/c5c2eaba230079662bc9be92e41b279492e78f58/examples/esp32/src/bin/ble_bas_peripheral.rs#L43).

It can be made `'static` by calling `mk_static` on the wifi "init" and then the controller itself can be made static but that's exactly what I want to avoid, because I want to tear down the BT stack once it is done.

[...]

PR #255, opened 2025-01-20
refactor: remove C from HostResources

lulf:
* Refactor API to remove C: Controller from HostResources
* Remove Clone + Copy on Stack and use it to hold the BleHost
* Remove all 'internal' lifetime of elements within channels. Guarantee their lifetime in types that is handed ownership of elements from channels.
* Simplify packet pool, removing the QoS flag.

Issue #252


ivmarkov:
@lulf Let me do a quick check if I can use your PR in `rs-matter-embassy` (i.e. that the stuff type-checks at least)

ivmarkov:
@lulf Type-checks in my code just fine! Huge improvement! (Yet to make it run of course.)

Thank you!

PR #256, opened 2025-01-20
Fixed a bug when setting the size of the `send`'s `TX_MTU`

hicklin:
The `ch1.send`'s `TX_MTU` generic const was set to the `PAYLOAD_LEN` which is always going to be wrong as the `encode` method adds 4 to 6 bytes at the start of the payload. This has been updated to `L2CAP_MTU`; the maximum allowed size.

### DIFF
--- a/examples/apps/src/ble_l2cap_central.rs
+++ b/examples/apps/src/ble_l2cap_central.rs
@@ -26,7 +26,7 @@ where
     } = stack.build();
 
     // NOTE: Modify this to match the address of the peripheral you want to connect to.
-    // Currently it matches the address used by the peripheral examples
+    // Currently, it matches the address used by the peripheral examples
     let target: Address = Address::random([0xff, 0x8f, 0x1a, 0x05, 0xe4, 0xff]);
 
     let config = ConnectConfig {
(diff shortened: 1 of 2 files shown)


lulf:
The payload len can be bigger than l2cap mtu, and it will be fragmented internally. 27 may be a bit confusing though, it should probably be set to a larger number to not be confused with the l2cap mtu.

lulf:
There are basically 3 different MTUs here:

1. The l2cap mtu, which is configured in the HostResources
2. The l2cap connection oriented channel MPS: this is configured in the L2capChannelConfig `mtu` field. This may be smaller or larger than the l2cap mtu. Fragmentation is done before sending to controller.
3. The 'tx buffer' passed to `send()`. This must be at least as big as the MPS, but may be bigger. Fragmentation to the MPS size is happening in the internal `send()`.

lulf:
@hicklin Maybe an alternative here would be to increase the PAYLOAD_LEN to some larger arbitrary value like 42?

lulf:
@hicklin Let's merge this for now then, and sorry for the misunderstanding!

PR #258, opened 2025-01-23
potential for cleanup

lure23:
Seems to me that these three are leftovers:

- `num_enum`
- `thiserror`
- `critical-section` (in that context)

The three are shown dimmed by my IDE (Rust Rover), which normally is an indication that they aren't used. However, I have only tested this locally by building one `esp32` example. Partly creating the PR to see if the change makes any tests fail (i.e. if they were indeed used).


lure23:
Placed `critical-section` back, since CI builds failed on it.

lulf (approved):

Issue #248, opened 2025-01-17
Proc-macros: useless conversion Clippy lint


ivmarkov:
Consider:

```rust
/// Matter service
#[gatt_service(uuid = MATTER_BLE_SERVICE_UUID16)]
struct MatterService {
    #[characteristic(uuid = Uuid::Uuid128(C1_CHARACTERISTIC_UUID.to_be_bytes()), write)]
    c1: External,
    #[characteristic(uuid = Uuid::Uuid128(C2_CHARACTERISTIC_UUID.to_be_bytes()), write, indicate)]
    c2: External,
}
```

... where, inside `rs-matter`, I have:
```rust
pub const MATTER_BLE_SERVICE_UUID16: u16 = 0xFFF6;
pub const C1_CHARACTERISTIC_UUID: u128 = 0x18EE2EF5263D4559959F4F9C429F9D11;
pub const C2_CHARACTERISTIC_UUID: u128 = 0x18EE2EF5263D4559959F4F9C429F9D12;
```

With `MATTER_BLE_SERVICE_UUID16` everything is great, because trouble's `Uuid` has `impl From<u16> for Uuid`.
With c1 and c2 life is not so great, because `Uuid` has neither `impl From<[u8; 16]> for Uuid`, nor `impl From<u128> for Uuid`.

Therefore, I _have to_ use the explicit `Uuid::Uuid128(C1_CHARACTERISTIC_UUID.to_be_bytes())`.
However, since the proc-macro generated code does internally do `.into()` on my `Uuid`, Clippy then kicks-in with:

```
warning: useless conversion to the same type: `trouble_host::prelude::Uuid`
  --> src/ble.rs:45:29
   |
45 |     #[characteristic(uuid = Uuid::Uuid128(C1_CHARACTERISTIC_UUID.to_be_bytes()), write)]
   |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ help: consider removing `.into()`: `Uuid::Uuid128(C1_CHARACTERISTIC_UUID.to_be_bytes())`
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
   = note: `#[warn(clippy::useless_conversion)]` on by default
```

I think we either need extra `From<???> for Uuid` impls, or this (and other) Clippy lints disabled in the proc-macros, or ideally both.

lure23:
[...]

PR #259, opened 2025-01-23
suggestion: equal processing for both 16- and 128-bit Uuids  // WIP

lure23:
Related to https://github.com/embassy-rs/trouble/issues/248 , I'd like to see both 16- and 128-bit Uuid processing logic to be alike. I think this is a good deed for maintainability and clarity.

They currently differ in a few places, and I'd like to discuss those within the PR. If this gets thumbs, could massage the PR to mergeable quality.


host-macros/src/uuid.rs
```diff
@@ -11,15 +11,13 @@ use quote::quote;
...
     Uuid16(u16),
-    Uuid128([u8; 16]),
+    Uuid128(u128)   //was: Uuid128([u8; 16]),
 }
 
 impl FromMeta for Uuid {
     fn from_string(value: &str) -> darling::Result<Self> {
         if let Ok(u) = uuid::Uuid::from_str(value) {
-            let mut bytes = u.as_bytes().to_owned();
-            bytes.reverse(); // Little-endian, as per Bluetooth spec
```
lure23, line 21: Note that here, endianess conversion is non-conditional. This seems like a bug to me, but those would be removed lines in the PR.

host/src/types/uuid.rs
```diff
@@ -9,7 +9,7 @@ use crate::codec::{Decode, Encode, Error, Type};
 #[derive(Debug, PartialEq, Clone)]
 pub enum Uuid {
     /// 16-bit UUID
-    Uuid16([u8; 2]),
+    Uuid16([u8; 2]),    // Q: why here as slice, when 'u16' in 'host-macros'?
```
lure23, line 12: Is there a reason why the inner contents are arrays here, but `u16` in `host-macros`?

Should we make them both the same way?

host/src/types/uuid.rs
```diff
@@ -27,7 +27,11 @@ impl Uuid {
     }
 
     /// Create a new 128-bit UUID.
-    pub const fn new_long(val: [u8; 16]) -> Self {
+    pub const fn new_long(val: u128) -> Self {
+        Self::Uuid128(val.to_le_bytes())
+    }
+        // tbd. this is an API change
+    pub const fn new_long_from_arr(val: [u8; 16]) -> Self {
```
lure23, line 34: There are two places (below, in the `tests`) that use this function, but it's also public.

Is there a need for both; can tests use long hex constants (instead of strings)?  How would building from arrays / strings be unified, perhaps using `From`, see #248

host/tests/gatt.rs
```diff
@@ -9,10 +9,10 @@ mod common;
 const CONNECTIONS_MAX: usize = 1;
 const L2CAP_CHANNELS_MAX: usize = 3;
 
-const SERVICE_UUID: Uuid = Uuid::new_long([
+const SERVICE_UUID: Uuid = Uuid::new_long_from_arr([
```
lure23, line 12: Another approach is:

```rust
const SERVICE_UUID: Uuid = Uuid::new_long(
   0x00001000b0cd11ec871fd45ddf138840
);
```

jamessizeland:
Looking good, make sure you check that the conversions are equivalent (maybe a test temporarily to show that?) or better if you think we've got an endienness bug?

lure23:
@jamessizeland Do you know why the CI build is failing? I have no idea what these are:

```
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Diff in /home/runner/work/trouble/trouble/host/src/types/uuid.rs:9:
 #[derive(Debug, PartialEq, Clone)]
 pub enum Uuid {
     /// 16-bit UUID
-    Uuid16([u8; 2]),    // Q: why here as slice, when 'u16' in 'host-macros'?
+    Uuid16([u8; 2]), // Q: why here as slice, when 'u16' in 'host-macros'?
     /// 128-bit UUID
     Uuid128([u8; 16]),
 }
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
Warning: can't set `imports_granularity = Module`, unstable features are only available in nightly channel.
Warning: can't set `group_imports = StdExternalCrate`, unstable features are only available in nightly channel.
```

host/src/types/uuid.rs
lure23: @lulf Would appreciate your comment on this.

host-macros/src/uuid.rs
alexmoon: No, this is correct. The `uuid` crate stores UUIDs in big-endian order, but the bluetooth spec requires UUIDs to be transmitted in little-endian order.

host/src/types/uuid.rs
alexmoon: `[u8; 2]` should be preferred because it has align of 1 instead of align of 2.

host/tests/gatt.rs
alexmoon: To match current behavior it should be reversed.

lulf:
This seems to have stalled, is there anything anyone can help out with? I'm wondering if it's better to track this feature request in a github issue for now (if it is still an issue). What do you think, @lure23?

lure23:
@lulf I'm coming back to Trouble eventually, but currently doing other work.  I don't mind if we close this issue - personally I prefer having a lesser number of active issues in my work. If there is lack of parity between 16 and 128-bit API's, it will re-surface at some point.

lulf:
@lure23 Great, feel free to re-open it when you've got the time !

PR #261, opened 2025-01-23
Allow getting and setting all attributes

### LINKED ISSUE #155: Expand get and set methods to allow getting and setting all attributes

petekubiak:
Implements generic get and set methods, allowing getting and setting of descriptors and characteristic properties.

This is a draft to get feedback on what you think of the implementation, I think some documentation and examples are needed before merging.

Resolves #155.


lulf (approved):
Looks good to me!

jamessizeland:
Do we need to assert that a writeable descriptor must also be named?

petekubiak:
> Do we need to assert that a writeable descriptor must also be named?

I don't think so, because setting the descriptor as writeable only affects whether it can be written by a connected device.
The user might not need to access the descriptor locally, so we don't want to have to force them to name it.

host/src/types/gatt_traits.rs
```diff
@@ -69,6 +69,7 @@ impl Primitive for f32 {}
 impl Primitive for f64 {}
 impl Primitive for BluetoothUuid16 {} // ok as this is just a NewType(u16)
 impl Primitive for &'_ str {}
+impl Primitive for &'_ [u8] {}
```
petekubiak, line 72: This impl follows the same flawed logic highlighted in #272 and will require the same fix (assuming I can find one 😅)

jamessizeland (approved):

host/src/types/gatt_traits.rs
petekubiak: Ok, we now have a solution for this so I can hopefully close this off soon.

lulf (approved):
No objection from me, thanks! Needs some fmt'ing.

PR #275, opened 2025-01-27
Add esp32 ble_scan example.

elrom:


elrom:
Interestingly, at least on my board (esp32s3), I'm not getting any results from the scan. @lulf  any thoughts?

lulf:
@elrom I've only tested it on nrf, so maybe an esp-specific thing yeah. I'll try to test this on my esp32c3 and see if I can find anything.

lulf:
@elrom Seems like there is someone who got it to work after making L2CAP_MTU configurable in the example, is you make that modification I'd be happy to merge this PR (ref https://github.com/embassy-rs/trouble/issues/288)

Meigs2:
Yes, all you have to do is pass in a const generic for the run method, `pub async fn run<C, const L2CAP_MTU: usize>(controller: C)` and use that const instead of the hardcoded one in the example. This will require modification to the other examples to pass in a value for `L2CAP_MTU` to get them to compile.

lulf:
Example exists now

PR #278, opened 2025-01-29
Proposal: Add MegaLinter and Cargo Clippy CI Cycle

ScottGibb:
# Summary
This PR is a suggestion for adding [MegaLinter](https://megalinter.io/latest/) to the project. MegaLinter provides a lot of linters that will auto format and fix various issues with the codebase. Currently I have it setup to lint all the different file types we use in Trouble. 

## Why

- Automatically check for formatting and adherence to cargo clippy suggestions
- keep formatting and code style consistent across the project using cargo and various other standardised linters.

## General workflow

Whenever a PR is raised the linters will run on the PR and check all the files are correctly formatted, links are correctly working etc. We also run cargo clippy on all our examples and crates to ensure they all meet appropriate standards. If a formatting change can be fixed, the workflow will create a PR into the existing PR with the apropriate changes. Thus the code can be reformatted on GitHub.

# Why all the files changes?

Most of these files are formatting changes, which have occurred from running cargo fmt on the project.

## Gotchas

Sadly Megalinters support for rust is limited (still available) and due to the structure of the project ive added a couple of extra jobs/stages in order to ensure full coverage. However in the future they will be adding better rust support.

# Implementation Issues regarding workflow minutes

If GitHub workflow minutes are a potential barrier, we can run this on a `self-hosted` Runner like a Pi5 or otherwise. Thus having no impact on embassy GitHub minutes. I believe the workflow only takes a couple of minutes at the moment anyway.

# Current Issues to sort out (before merge)

- [ ] Apache-Nimble example: Im currently unable to run/build this example and thus I cant get clippy to run either. I believe it has something to do with this issue: https://github.com/benbrittain/apache-nimble-sys/issues/6

- [ ] esp32 examples: Im a bit confused on how I can install the esp extensa toolchain to clippy the examples, these examples are currently unclippied:
         - esp32
         - esp32s3

- [ ] Add MegaLinter instructions/description to main readme
- [ ] Add PAT Token so megalinter can create PRs with format fixes


lulf (commented):
TIL megalinter :) Thanks for the PR

As with all these tools, I have a hard requirement that it should be possible to run it locally and being able to correct things (ideally automatically like cargo fmt) before creating a PR. 

Also, from looking at the configuration, it seems like we could get the same benefit (apart from shell, toml validation) from adding the clippy checks to ci.sh?

ScottGibb:
So you can run this from command line too!! I haven't added the command to call/install to the readme section yet, but will do when I've got access to my laptop. 

The real benefit of this, over adding to the ci.sh, is we don't have to maintain any of it as the setup is done by mega linter. We just have to maintain a config file and a GHA file.


As for the clippy bits adding to ci.sh, we could potentially but I'm hoping with future improvements to mega-linter we can do it all from mega-linter.

I think there is a way to get mega linters rust clippy to run more than once with more arguments but haven't dug around deep enough yet.

lulf:
> So you can run this from command line too!! I haven't added the command to call/install to the readme section yet, but will do when I've got access to my laptop.
> 

Okok, let's give it a try :) Thanks for introducing some new stuff and processes!

> The real benefit of this, over adding to the ci.sh, is we don't have to maintain any of it as the setup is done by mega linter. We just have to maintain a config file and a GHA file.
> 

I just want to run 'one thing' to do it all, so in that case please add some way for ci.sh to run the megalinter as well (like, if it detects the CLI tool). Is it possible to install that in the 'prereqs' of the GHA and then run the tool as part of ci.sh instead of using a separate job/action?

The reason is that I really don't like creating too many workflows, especially since there is a shared quota on the embassy organization, and that it can easily get out of hand if developers can't run it all locally in a simple way.

I would actually like to run all integration and example tests in ci.sh as well, but can't because it could then only run on machines that has the hardware.  But at least with pure software, we can run it wherever we want!

lulf:
@ScottGibb You can install it on a mac like this:

```
cargo install --git https://github.com/embassy-rs/cargo-batch --branch cargo-batch-1.85.0 cargo --bin cargo-batch --locked
```

ScottGibb:
End of Day Notes:

- Use shell script arguments for -- fix to apply megalinter fixes and cargo fmt fixes locally
- Make sure ci.sh fails if the fmt or clippy shows warnings

lulf:
Is this still something worth doing, or should it be closed until there is more time to focus on it @ScottGibb ?

ScottGibb:
Hey @lulf it may be worth closing till I have time to look at it. Sorry i've not had a look at it yet. Its on the open source priority list :)

lulf:
No worries, I'm just doing some 'house cleaning' :)

PR #289, opened 2025-02-19
Split from_gatt and to_gatt into separate traits and implement for &'static str

### LINKED ISSUE #272: Impl of FixedGattValue for &str seems wrong
I was attempting to send a string as a characteristic with the following code:
```rust
#[gatt_service(uuid = DEVICE_INFO_SERVICE_UUID)]
struct DeviceInfoService {
    #[characteristic(uuid = "00002a25-0000-1000-8000-00805f9b34fb", value = "1337", read)]
    serial_number: &'static str,
}
```

I could see the characteristic but it had an odd nonsense value that was not valid utf8

Looking at the code for the `impl<T: Primitive> FixedGattValue for T` it seems that `SIZE` is set as ` mem::size_of::<Self>();`  and in the case of `T = &str`  this wont actually be the size of the str slice, just the size of the reference.

Having realized this I switched to trying to use `heapless::String` (as shown below) and that worked as expected. 

```rust
#[gatt_service(uuid = DEVICE_INFO_SERVICE_UUID)]
struct DeviceInfoService {
    #[characteristic(uuid = "00002a25-0000-1000-8000-00805f9b34fb", value = heapless::String::from_str("1337").unwrap(), read)]
    serial_number: heapless::String<8>,
}
```

petekubiak:
The implementation of `GattValue` for `&'_ str` was incorrect, and was returning the reference's memory on the stack rather than the slice it was pointing to. Furthermore, there were issues linking lifetimes of references as gatt values to their &[u8] equivalent (see #272 for more details)

This PR splits `GattValue` into separate traits `FromGatt` and `ToGatt` where `FromGatt` requires that `ToGatt` is implemented. This allows us to implement `to_gatt` on `&'static str` and other static references without having to implement `from_gatt`, which would potentially be unsafe.

Fixes #272.

### DIFF
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -29,7 +29,7 @@ mod fmt;
 #[cfg(not(any(feature = "central", feature = "peripheral")))]
 compile_error!("Must enable at least one of the `central` or `peripheral` features");
 
-mod att;
+pub mod att;
 #[cfg(feature = "central")]
 pub mod central;
 mod channel_manager;
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -97,7 +97,7 @@ pub mod prelude {
     #[cfg(feature = "scan")]
     pub use crate::scan::*;
     #[cfg(feature = "gatt")]
-    pub use crate::types::gatt_traits::{FixedGattValue, GattValue};
+    pub use crate::types::gatt_traits::FixedGattValue;
     pub use crate::Address;
 }
 
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -97,7 +97,7 @@ pub mod prelude {
     #[cfg(feature = "scan")]
     pub use crate::scan::*;
     #[cfg(feature = "gatt")]
-    pub use crate::types::gatt_traits::FixedGattValue;
+    pub use crate::types::gatt_traits::{FixedGattValue, FromGatt, ToGatt};
     pub use crate::Address;
 }
 
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -89,7 +89,7 @@ pub mod prelude {
     pub use crate::gap::*;
     #[cfg(feature = "gatt")]
     pub use crate::gatt::*;
-    pub use crate::host::{ControlRunner, HostMetrics, Runner, RxRunner, TxRunner};
+    pub use crate::host::{ControlRunner, EventHandler, HostMetrics, Runner, RxRunner, TxRunner};
     pub use crate::l2cap::*;
     pub use crate::packet_pool::PacketPool;
     #[cfg(feature = "peripheral")]
--- a/host/src/types/gatt_traits.rs
+++ b/host/src/types/gatt_traits.rs
@@ -15,7 +15,7 @@ pub enum FromGattError {
 
 /// Trait to allow conversion of a fixed size type to and from a byte slice
 #[allow(private_bounds)]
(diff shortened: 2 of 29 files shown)


jamessizeland:
Just a thought, should the trait be IntoGatt rather than ToGatt? I'm not sure which is more idiomatic.

host/src/lib.rs
```diff
@@ -97,7 +97,7 @@ pub mod prelude {
     #[cfg(feature = "scan")]
     pub use crate::scan::*;
     #[cfg(feature = "gatt")]
-    pub use crate::types::gatt_traits::{FixedGattValue, GattValue};
```
jamessizeland, line 100: We're no longer exposing the traits in the prelude, is this the plan?

petekubiak:
I wondered this, and then thought that if the trait was called IntoGatt, the function should also be called into_gatt...

host/src/lib.rs
petekubiak: Nope! That was me on autopilot xD

jscatena88:
In reference to the trait names there is potentially an argument to be made that it should be `AsGatt` if we think the conversions will always be fairly cheap. `To` implies that is an expensive/slow conversion: https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv

This is a nit though so don't feel obligated at all

host/src/types/gatt_traits.rs
```diff
@@ -138,22 +143,38 @@ impl<const N: usize> GattValue for [u8; N] {
...
 }
+
+impl ToGatt for &'static str {
+    const MIN_SIZE: usize = 0;
+    const MAX_SIZE: usize = usize::MAX;
+
+    fn to_gatt(&self) -> &[u8] {
+        self.as_bytes()
+    }
+}
```
jscatena88, line 180: Does this have to be static? I'd imagine almost any use case would have it be static but I wonder if you can get a little extra flexibility by using a generic lifetime. This should work the same for a &`static str I believe while also allowing non-static `&str`s to be used assuming the borrow checker agrees.

```suggestion
impl<'a> ToGatt for &'a str {
    const MIN_SIZE: usize = 0;
    const MAX_SIZE: usize = usize::MAX;

    fn to_gatt(&self) -> &'a [u8] {
        self.as_bytes()
    }
}
```

jscatena88:
Overall looks like a great improvement! I don't have time at this exact moment to test on HW or on the project I was working on but I see no reason it wouldn't all work

jscatena88 (approved):

host/src/types/gatt_traits.rs
lulf: See the discussion here https://github.com/embassy-rs/trouble/issues/272 @jscatena88

lulf (approved):
Thanks for fixing this, looks great!

petekubiak:
Just to mention, I'm fully on board with moving to AsGatt instead, I think there's a strong argument there. I'll raise a PR for it.

host/src/types/gatt_traits.rs
jscatena88: It looks like the complications there were from implementing `from_gatt` not `to_gatt`. I'm still somewhat confident that the code I wrote above should work actually. It is passing tests locally on my laptop (except for one test that seems to be expecting connected hardware? `gatt_client_server`). But either way not a big deal

PR #298, opened 2025-02-25
chore: move to rust edition 2024

lulf:


jamessizeland:
Nice! I started this and then got distracted after seeing the number of fixes to do! Well done!

jamessizeland:
This looks like less than I saw from `cargo fix --edition` or maybe I'm imagining it.

PR #301, opened 2025-02-26
Remove invalid field from FindTypeValue response

### LINKED ISSUE #107: Potential issue in GATT processing using Windows LE explorer
Using Windows LE explorer is with trouble + esp32 is reported to hang. See https://github.com/esp-rs/esp-hal/pull/1971#issuecomment-2311760670 for a packet capture. Attempt to reproduce with other HCI controllers as well to see if it's esp-specific.

timokroeger:
This incorrect field prevented the WinRT Bluetooth API to work with TrouBLE devices. Fixes #107.


lulf (approved):

lulf:
Thanks for reproducing and fixing this!

PR #302, opened 2025-02-26
Update the GATT client to allow use of `Write Without Response` characteristics

Meigs2:
# Summary
Some BLE devices expose the attribute "Write Without Response" on characteristics. Currently, `trouble` only supports writing *with* responses. This PR extends the GATT client to support Write Without Response.

## Implementation
[According to the BLE Spec regarding Write Without Response](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/attribute-protocol--att-.html), the end of section `3.4.5.3` reads:

> No ATT_ERROR_RSP or ATT_WRITE_RSP PDUs shall be sent in response to this command. If the server cannot write this attribute for any reason the command shall be ignored

Therefore, we should be able to re-use existing logic in the `request` method for the GATT client to send the message, and not wait on a response from the `response_channel`.

I've added a new method, `write_characteristic_without_response` which invokes this new `Write` command process

## Testing

I've tested this on an Acaia Lunar 2021 which exposes a Write Without Response characteristic, writing to this before would cause an `WRITE_NOT_PERMITTED = 0x03` error code from the host layer. Utilizing the new method allows me keep a long-lived connection open with the scale. This was tested on an ESP32C3 dev board.

### DIFF
--- a/host/src/att.rs
+++ b/host/src/att.rs
@@ -768,7 +768,7 @@ impl<'d> AttCmd<'d> {
         let mut w = WriteCursor::new(dest);
         match self {
             Self::Write { handle, data } => {
-                w.write(ATT_WRITE_REQ)?;
+                w.write(ATT_WRITE_CMD)?;
                 w.write(*handle)?;
                 w.append(data)?;
             }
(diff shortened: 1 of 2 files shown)


lulf (approved):
LGTM, thanks for fixing!

PR #304, opened 2025-03-02
Add support for advertised 16 and 128 bit service uuid

Easyoakland:
Add support for 16 bit complete, 16 bit with data, and 128 bit complete advertised service uuids.

I've only tested the 16bit complete service, but I assume the others works as well.

I'm not particularly familiar with Bluetooth, so I don't know if the same approach would work for the "Incomplete List" variants since I can't find where the format is specified.


lulf (approved):
Looks good, agreed with changing to use the arrays in the API to enforce the correct uuid type :+1:

Issue #306, opened 2025-03-02
Allow splitting L2capChannel into a read half and a write half.


Dominaezzz:
I'm implementing IPSP/[RFC7668](https://datatracker.ietf.org/doc/html/rfc7668) on top of trouble.

To do so I need to be able to read and write data (IPv6 packets) concurrently.

I'm currently doing something like this:
```Rust
let result = select(
    // Wait for network stack/embassy-net to transmit a packet.
    tx.tx_buf(),
    // Wait for packet to come in the L2CAP channel.
    self.channel.receive(&self.stack, &mut buffer),
).await;
```

Whilst it works for a POC, the problem with it is `channel.receive` isn't cancellation friendly.
If `tx.tx_buf()` returns while `self.channel.receive` is reassembling PDUs, this received data will be lost when it's cancelled.

I don't specifically need to split the channel, I just need some way to read and write concurrently.

lulf:
You can already .clone() an l2cap channel, is that not what you're looking for? Although now that you've mentioned it, I think I'd like to change the api to use a split() in the future.

Dominaezzz:
> You can already .clone() an l2cap channel, is that not what you're looking for?

Whoops, didn't notice that. Yeah that would work.

I'm curious what behaviour to expect if two clones read (or write) at the same time. Will I get garbage or will each come randomly get a full packet?

Splitting would make this more intuitive I think

PR #307, opened 2025-03-03
refactor: provide separate writer and reader instead of clone

lulf:
Rather than supporting clone() of l2cap channels, provide a split() funciton for creating separate write and read endpoints, since this enforces that only a single user can write to or read from at a time.


lulf:
Fixes #306

Dominaezzz:
Looks good! It'd be nice to also have a way to undo the split, like a `L2capChannel::merge(rx, tx)` kinda thing.

PR #309, opened 2025-03-03
fix: notification issues

### LINKED ISSUE #308: Notification issues
I've been having some trouble (ha!) getting notifications to work correctly within my project: some of them work as expected, while others aren't getting sent out. This project is a peripheral/server that supports 18 characteristics that can be configured to 'notify'. After doing some digging, I discovered the source of the problem to be within the `NotificationTable` of the `AttributeServer`... Specifically, there appear to be two separate issues:

1. **Notification Table size** : The notification table size is fixed to `MAX_NOTIFICATIONS`, which is currently 4. Each time a CCCD is received from a client, this table is updated to reflect whether notifications are enabled for this particular CCCD handle (`AttributeServer::set_notify()`). This is then used as a filter to determine when notifications will actually be sent out (`AttributeServer::should_notify()`), so it's not surprising that some of my notifications aren't working: there simply aren't enough slots!  So how many table entries should there be? We could pretty easily add a CCCD count in the `#[gatt_server]` macro that works exactly like the way attributes are added up for all the registered services. But each notification table entry also stores a connection handle to track CCCD handles for more than one connected client... So the total possible number of CCCD/notification table entries is actually the above CCCD count multiplied by the maximum number of connections (`CONNECTIONS_MAX`). This would mean the user would have to pass `CONNECTIONS_MAX` in as an argument; something like `#[gatt_server(connections_max = CONNECTIONS_MAX)]` in addition to using it in `HostResources`. Maybe a little awkward, but not the end of the world. But about those connection handles...
2. **Connection state is untracked in Notification Table** : As mentioned above, a client obtains or releases an entry within the notification table when they send a CCCD that indicates their notification preference for that particular CCCD handle. This stores the connection handle for that client as well. The problem is that these entries are never released if the client disconnects. Even if the same client reconnects, the connection handle is likely to be different (in testing it appears to monotonically increase on every new connection), meaning they will need to claim new slots in the notification table, and the old ones are just taking up space. Eventually, when the table becomes full, notifications will just stop working altogether. So, in addition to the connection handle we need some tracking of connection state for each client: could just be a `is_connected: bool` that is set when a client opens a connection, and is cleared when they disconnect, freeing up that entry to be used by the next connecting client. I'd prefer to track the actual peer address instead of the connection handle for each table entry, because then at least the server could remember a previously connected client's configuration, which is a step towards fulfilling [Vol 3 Part G 3.3.3.3](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html#UUID-70d11f51-12cd-57a4-184a-fd8a4e0283f9)'s requirement that the CCC "shall be persistent across connections for bonded devices". At some point you could optionally provide the `AttributeServer` a blob of `embedded-storage` to maintain this information across reset/power-cycles as well, but that's beyond the scope of this particular issue.

To address both 1 & 2, I have a few changes I'd like to propose:
1. **GATT Server macro** : add two new attributes, `connections_max` (required) and `cccd_table_size` (optional). If not provided, `cccd_table_size` would be summed up across all of the services just like the `ATTRIBUTE_COUNT`. This is going to add two new `const` generic parameters to the `AttributeServer`, but `trouble` users should never see them, so this should be OK.
2. **GattConnection** : I'd like to add a new higher-level connection type the user would interact with (instead of the lower-level `Connection`) when utilizing the server-side functionality provided by `GattData` and `GattEvent`: it is effectively just a wrapper around a `Connection` and `dyn` reference to a `DynamicAttributeServer`. The intended usage is to create one of these from the raw `Connection` produced after advertising completes successfully that exists for the lifetime of the connection to the client. In the constructor, a call could be made to find a spot for the client within the `NotificationTable` (`AttributeServer::connect(&Connection)`), and then when this object is dropped there is a similar call to `AttributeServer::disconnect(&Connection)` to update its `NotificationTable` state. As the user's high-level interface for a GATT connection, the `GattConnection` would have a `next()` method just like `Connection`: it would probably make sense to just move this functionality out of the lower-level `Connection` type instead of calling into it, making these `GattConnectionEvent`s. Since the `GattConnection` already bundles a reference to a `DynamicAttributeServer`, you could remove either `Connection` or `Server` arguments in a few user-facing places, which makes the API a little nicer to work with (`Characteristic::notify()` and `GattData::process()`, for example), and internally you could just store a `GattConnection` within the `GattData` and `GattEvent` types instead of in separate `Connection` and `DynamicAttributeServer` fields.
3. **Rename `NotificationTable` to `CccdTables`** : This would eventually also store indication preferences as well, so each `CccdTable` (one per client) entry would include a `CCCD` value set with the appropriate `CCCDFlag`s.

I'll be submitting a PR shortly that addresses all of the above, and that I've tested as working correctly for multiple simultaneously connected clients supporting as many CCCDs as the `AttributeTable` has got, across a number of connection/disconnection events. This is my first foray into `trouble`, so forgive me if something doesn't look quite right! If there is a more appropriate solution, I'd also like to know about that as well as I am still learning the ways of Rust.. `trouble` is a very cool project, many thanks to all who have contributed to it!

chrissuozzo:
**Issue**
Closes #308

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -10,7 +10,7 @@ const CONNECTIONS_MAX: usize = 1;
 const L2CAP_CHANNELS_MAX: usize = 2; // Signal + att
 
 // GATT Server definition
-#[gatt_server(connections_max = CONNECTIONS_MAX)]
+#[gatt_server]
 struct Server {
     battery_service: BatteryService,
 }
@@ -109,48 +109,33 @@ async fn gatt_events_task(server: &Server<'_>, conn: &GattConnection<'_, '_>) ->
                 info!("[gatt] disconnected: {:?}", reason);
                 break;
             }
-            GattConnectionEvent::Gatt { data } => {
-                // We can choose to handle event directly without an attribute table
-                // let req = data.incoming();
-                // ..
-                // data.reply(Ok(AttRsp::Error { .. }))
-
-                // But to simplify things, process it in the GATT server that handles
-                // the protocol details
-                match data.process().await {
-                    // Server processing emits
-                    Ok(Some(event)) => {
-                        match &event {
-                            GattEvent::Read(event) => {
-                                if event.handle() == level.handle {
-                                    let value = server.get(&level);
-                                    info!("[gatt] Read Event to Level Characteristic: {:?}", value);
-                                }
-                            }
-                            GattEvent::Write(event) => {
-                                if event.handle() == level.handle {
-                                    info!("[gatt] Write Event to Level Characteristic: {:?}", event.data());
-                                }
+            GattConnectionEvent::Gatt { event } => match event {
+                Ok(event) => {
+                    match &event {
+                        GattEvent::Read(event) => {
+                            if event.handle() == level.handle {
+                                let value = server.get(&level);
+                                info!("[gatt] Read Event to Level Characteristic: {:?}", value);
                             }
                         }
-
-                        // This step is also performed at drop(), but writing it explicitly is necessary
-                        // in order to ensure reply is sent.
-                        match event.accept() {
-                            Ok(reply) => {
-                                reply.send().await;
-                            }
-                            Err(e) => {
-                                warn!("[gatt] error sending response: {:?}", e);
+                        GattEvent::Write(event) => {
+                            if event.handle() == level.handle {
+                                info!("[gatt] Write Event to Level Characteristic: {:?}", event.data());
                             }
                         }
                     }
-                    Ok(_) => {}
-                    Err(e) => {
-                        warn!("[gatt] error processing event: {:?}", e);
+
+                    // This step is also performed at drop(), but writing it explicitly is necessary
+                    // in order to ensure reply is sent.
+                    match event.accept() {
+                        Ok(reply) => {
+                            reply.send().await;
+                        }
+                        Err(e) => warn!("[gatt] error sending response: {:?}", e),
                     }
                 }
-            }
+                Err(e) => warn!("[gatt] error processing event: {:?}", e),
+            },
         }
     }
     info!("[gatt] task finished");
@@ -182,8 +167,7 @@ async fn advertise<'a, 'b, C: Controller>(
         )
         .await?;
     info!("[adv] advertising");
-    let conn = advertiser.accept().await?;
-    let conn = GattConnection::new(conn, server);
+    let conn = advertiser.accept().await?.with_attribute_server(server)?;
     info!("[adv] connection established");
     Ok(conn)
 }
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -167,7 +167,7 @@ async fn advertise<'a, 'b, C: Controller>(
     AdStructure::encode_slice(
         &[
             AdStructure::Flags(LE_GENERAL_DISCOVERABLE | BR_EDR_NOT_SUPPORTED),
-            AdStructure::ServiceUuids16(&[Uuid::Uuid16([0x0f, 0x18])]),
+            AdStructure::ServiceUuids16(&[[0x0f, 0x18]]),
             AdStructure::CompleteLocalName(name.as_bytes()),
         ],
         &mut advertiser_data[..],
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -10,7 +10,7 @@ const CONNECTIONS_MAX: usize = 1;
 const L2CAP_CHANNELS_MAX: usize = 2; // Signal + att
 
 // GATT Server definition
-#[gatt_server]
+#[gatt_server(connections_max = CONNECTIONS_MAX)]
 struct Server {
     battery_service: BatteryService,
 }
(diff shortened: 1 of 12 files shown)


examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -10,7 +10,7 @@ const CONNECTIONS_MAX: usize = 1;
 const L2CAP_CHANNELS_MAX: usize = 2; // Signal + att
 
 // GATT Server definition
-#[gatt_server]
+#[gatt_server(connections_max = CONNECTIONS_MAX)]
```
lulf, line 13: I think we should have a default of '1' in the macro, which covers most of the common 'personal accessory' use cases

host/src/connection.rs
```diff
@@ -191,24 +172,19 @@ impl<'stack> Connection<'stack> {
...
     #[cfg(feature = "gatt")]
-    pub async fn next(&self) -> ConnectionEvent<'stack> {
-        match self.manager.next(self.index).await {
-            ConnectionEventData::Disconnected { reason } => ConnectionEvent::Disconnected { reason },
-            ConnectionEventData::Gatt { data } => ConnectionEvent::Gatt {
-                data: crate::gatt::GattData::new(data, self.clone()),
-            },
-        }
+    pub(crate) async fn next(&self) -> ConnectionEventData {
+        self.manager.next(self.index).await
```
lulf, line 187: I'd like to keep ConnectionEvent and ConnectionEventData as is, because:

1. There is a use of reading connection events for non-GATT applications to discover disconnections.
2. We currently support the use case of handling GATT data without relying on the AttributeServer, which would not be possible anymore with this change. Using the attribute server is a trade-off between convenience and memory usage.

host/src/connection.rs
```diff
@@ -282,4 +258,13 @@ impl<'stack> Connection<'stack> {
             Err(e) => Err(e),
         }
     }
+
+    /// Transform BLE connection into a `GattConnection`
+    #[cfg(feature = "gatt")]
+    pub fn to_gatt<'values, 'server, M: RawMutex, const AT: usize, const CT: usize, const CN: usize>(
```
lulf, line 264: I'd like to rename this `with_attribute_server` to signify that you want to handle this connection with an attribute server.

host/src/gatt.rs
```diff
@@ -16,29 +16,84 @@ use heapless::Vec;
...
+    fn drop(&mut self) {
+        trace!("[gatt {}] disconnecting from server", self.connection.handle().raw());
+        self.server.disconnect(&self.connection);
+    }
+}
+
+impl<'stack, 'server> GattConnection<'stack, 'server> {
+    /// Creates a GATT connection from the given BLE connection and `AttributeServer`:
+    /// this will register the client within the server's CCCD table.
+    pub fn new<'values, M: RawMutex, const AT: usize, const CT: usize, const CN: usize>(
```
lulf, line 57: Make this pub(crate), to ensure only a single API to do the same thing.

host/src/gatt.rs
```diff
@@ -60,8 +115,9 @@ impl<'stack> GattData<'stack> {
 
     /// Respond directly to request.
     pub async fn reply(self, rsp: AttRsp<'_>) -> Result<(), Error> {
-        let pdu = send(&self.connection, AttServer::Response(rsp))?;
-        self.connection.send(pdu).await;
+        let connection = self.connection.raw();
+        let pdu = send(connection, AttServer::Response(rsp))?;
+        connection.send(pdu).await;
```
lulf, line 120: Now that you have to create a GattConnection, and therefore an AttributeServer, this API is kind of pointless, because the purpose of this was that you didn't have to create an AttributeServer in order to use it.

host/src/attribute_server.rs
```diff
@@ -1,31 +1,202 @@
...
+            false
+        })
+    }
 }
 
 /// A GATT server capable of processing the GATT protocol using the provided table of attributes.
-pub struct AttributeServer<'values, M: RawMutex, const MAX: usize> {
-    pub(crate) table: AttributeTable<'values, M, MAX>,
-    pub(crate) notification: Mutex<M, RefCell<NotificationTable<MAX_NOTIFICATIONS>>>,
+pub struct AttributeServer<'values, M: RawMutex, const AT: usize, const CT: usize, const CN: usize> {
```
lulf, line 189: I'd rather see some fully spelled names for these, or at the minimum some docs explaining what they mean.

host/src/gatt.rs
```diff
@@ -16,29 +16,84 @@ use heapless::Vec;
...
+    },
+    /// GATT event.
+    Gatt {
+        /// The event that was returned
+        data: crate::gatt::GattData<'stack, 'server, 'conn>,
+    },
+}
+
+/// Used to manage a GATT connection with a client.
+pub struct GattConnection<'stack, 'server> {
```
lulf, line 42: Since this is specific to gatt servers and not clients, I'm wondering if it should be named ServerConnection (so you can use gatt::ServerConnection) or GattServerConnection.

lulf (changes requested):
Thank you for the PR! I love the idea of introducing a GattConnection type for simplifying the API. I've left some comments inline on what changes I'd like to see. 

I think in general that we should the existing GattData, GattEvent, but move the attribute server specific 'process()' into the GattConnection and/or GattConnectionEvent, to keep supporting the 'direct reply' use case without the attribute server.

host/src/gatt.rs
chrissuozzo: Possibly? The `GattData` and `GattEvent` types are also only used with a server, but also have no indication within their names that this is the case.

lulf: Yeah, good point. Im fine leaving this as is for now.

lulf (approved):
I think this looks really good, thanks again @chrissuozzo!

PR #312, opened 2025-03-08
default cccd to 0 if value missing

### LINKED ISSUE #305: BleHost(OutOfMemory)
I'm trying to subscribe to a heart-rate monitor's characteristic.

I'm using an esp32 on the 54cd0465ca497495b5890d896fc1f048f0a0f1ee revision. It doesn't look like there's an update related to my issue, so it shouldn't matter.

The relevant (simplified) code is:
```rs
let connection = match central.connect(config).await.unwrap();
let client = GattClient::<_, 10, { consts::BLE_L2CAP_MTU }>::new(stack, &connection).await.unwrap();
    let ((), ()) = join(async { client.task().await.unwrap() }, async {
        let services = client.services_by_uuid(&consts::BLE_SERIVCE_UUID_16).await.unwrap();
        assert_eq!(services.len(), 1, "todo: != 1");
        let service = services.first().unwrap();

        let characteristic: Characteristic<heapless::Vec<u8, 8>> = client
            .characteristic_by_uuid(service, &consts::BLE_CHARACTERISTIC_UUID_LONG)
            .await
            .unwrap(); // <-------
        let mut listener = client
            .subscribe(&characteristic, false)
            .await
            .unwrap();
    })
    .await;
```

I'm getting an error at the above where `<------` is indicated with `BleHost(OutOfMemory)`.

It looks like this originates here
https://github.com/embassy-rs/trouble/blob/92841fcc8bb986368456239a7bbf25be2f709ec1/host/src/gatt.rs#L622

I'm not sure what this code is doing or what a cccd is, but it seems strange that converting a byte slice to a u16 fails with OutOfMemory. There's not enough information here to determine what is actually broken, which makes the error unhelpful.

If the issue is that a buffer needs to be changed, what is that buffer and how is it increased?

If it helps, I've added a log with debug prints below.

<details><summary>log with added prints</summary>

```
INFO - services [ServiceHandle { start: 40, end: 65535, uuid: Uuid16([13, 24]) }]
INFO - [C:\path\trouble\host\src\gatt.rs:531] pdu request start
INFO - [C:\path\trouble\host\src\gatt.rs:783] att: Server(Response(ReadByType { it: ReadByTypeIter { item_len: 21, cursor: ReadCursor { pos: 2, data: [9, 21, 41, 0, 16, 42, 0, 251, 52, 155, 95, 128, 0, 0, 128, 0, 16, 0, 0, 55, 42, 0, 0] } } }))
INFO - [C:\path\trouble\host\src\gatt.rs:536] match start on ReadByType { it: ReadByTypeIter { item_len: 15, cursor: ReadCursor { pos: 2, data: [9, 15, 29, 0, 10, 2a, 0, fb, 34, 9b, 5f, 80, 0, 0, 80, 0, 10, 0, 0, 37, 2a, 0, 0] } } }
INFO - [C:\path\trouble\host\src\att.rs:400] cursor: ReadCursor { pos: 2, data: [9, 21, 41, 0, 16, 42, 0, 251, 52, 155, 95, 128, 0, 0, 128, 0, 16, 0, 0, 55, 42, 0, 0] }
INFO - [C:\path\trouble\host\src\att.rs:403] handle,item=(29, [10, 2a, 0, fb, 34, 9b, 5f, 80, 0, 0, 80, 0, 10, 0, 0, 37, 2a, 0, 0])
INFO - [C:\path\trouble\host\src\gatt.rs:544] item.len() >=5, decoding declaration
INFO - [C:\path\trouble\host\src\gatt.rs:551] decoded declaration
INFO - [C:\path\trouble\host\src\gatt.rs:557] calling get_characteristic_cccd
INFO - [C:\path\trouble\host\src\gatt.rs:783] att: Server(Response(ReadByType { it: ReadByTypeIter { item_len: 2, cursor: ReadCursor { pos: 2, data: [9, 2, 43, 0] } } }))
INFO - [C:\path\trouble\host\src\att.rs:400] cursor: ReadCursor { pos: 2, data: [9, 2, 43, 0] }
INFO - [C:\path\trouble\host\src\att.rs:403] handle,item=(2b, [])
INFO - [C:\path\trouble\host\src\gatt.rs:601] item: []
[31m

```
</details>

Easyoakland:
Fixes #305.

Since the spec says "The Client Characteristic Configuration declaration is an optional characteristic descriptor" and "The default value for the Client Characteristic Configuration descriptor value shall be 0x0000." and the peripheral in that issue works with this pr, I think this is correct.

I also changed the error on incorrect sized values (not 0 or 2) to be `InvalidValue` since the previous error of `OutOfMemroy` doesn't make sense.

### DIFF
--- a/host/src/gatt.rs
+++ b/host/src/gatt.rs
@@ -701,6 +701,7 @@ impl<'reference, C: Controller, const MAX_SERVICES: usize, const L2CAP_MTU: usiz
             AttRsp::ReadByType { mut it } => {
                 if let Some(Ok((handle, item))) = it.next() {
                     // As defined in the bluetooth spec [3.3.3.3. Client Characteristic Configuration](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html#UUID-09487be3-178b-eeca-f49f-f783e8d462f6)
+                    // "The Client Characteristic Configuration declaration is an optional characteristic descriptor" and
                     // "The default value for the Client Characteristic Configuration descriptor value shall be 0x0000."
                     if item.is_empty() {
                         Ok((handle, CCCD(0)))
--- a/host/src/gatt.rs
+++ b/host/src/gatt.rs
@@ -700,10 +700,16 @@ impl<'reference, C: Controller, const MAX_SERVICES: usize, const L2CAP_MTU: usiz
         match Self::response(response.pdu.as_ref())? {
             AttRsp::ReadByType { mut it } => {
                 if let Some(Ok((handle, item))) = it.next() {
-                    Ok((
-                        handle,
-                        CCCD(u16::from_le_bytes(item.try_into().map_err(|_| Error::OutOfMemory)?)),
-                    ))
+                    // As defined in the bluetooth spec [3.3.3.3. Client Characteristic Configuration](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html#UUID-09487be3-178b-eeca-f49f-f783e8d462f6)
+                    // "The default value for the Client Characteristic Configuration descriptor value shall be 0x0000."
+                    if item.is_empty() {
+                        Ok((handle, CCCD(0)))
+                    } else {
+                        Ok((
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks for checking and fixing!

PR #313, opened 2025-03-08
Enable waiting until a scan cancel command is completed

Easyoakland:
By waiting until a `ScanSession` cancel command is processed this prevents disallowed commands errors if future commands are sent to quickly after a `ScanSession` is dropped

In my particular case I first scan for a particular advertised service uuid, then attempt to connect to it.

When I do this on the ESP32 I get `BleHost(Hci(Command Disallowed))` if I don't insert a `Timer::after_micros(1000).await` between the dropping the output of `scanner.scan` and calling `central.connect`. I'm pretty sure this is because the `cancel` sent by dropping the session takes time to finish.

After this change I can now do `session.wait_idle().await` instead, which is more reliable and faster.

### DIFF
--- a/host/src/scan.rs
+++ b/host/src/scan.rs
@@ -136,10 +136,25 @@ impl<'d, C: Controller> Scanner<'d, C> {
 }
 
 /// Handle to an active advertiser which can accept connections.
+///
+/// # Warning
+/// If `ScanSession::wait_idle` is not used to drop the `ScanSession` then future attempts to send commands over the HCI (such as `central.connect`) may error for an indeterminate time later.
 pub struct ScanSession<'d, const EXTENDED: bool> {
     command_state: &'d CommandState<bool>,
     deadline: Option<Instant>,
     done: bool,
+}
+
+impl<const EXTENDED: bool> ScanSession<'_, EXTENDED> {
+    /// Cancel the `ScanSession` and wait until commands can be sent again.
+    ///
+    /// # Warning
+    /// If this is not used to drop the `ScanSession` then future attempts to send commands over the HCI (such as `central.connect`) may error for an indeterminate time later.
+    pub async fn wait_idle(self) {
+        let Self { command_state, .. } = self;
+        drop(self); // drop self which sends the cancel request.
+        command_state.wait_idle().await // wait until commands can be sent again.
+    }
 }
 
 impl<const EXTENDED: bool> Drop for ScanSession<'_, EXTENDED> {


lulf (commented):
I'm wondering if it would be more appropriate to add wait_idle() inside connect/connect_ext? This way it would be handled transparently.

lulf:
Actually this should already be handled here https://github.com/embassy-rs/trouble/blob/main/host/src/central.rs#L35 

The request() will wait until the command state is idle, which is set after the controller has ack'ed the cancel.

Easyoakland:
It seems more debugging is needed. The state is idle when connect is called, but the scan continues even after the request.
```
INFO - connecting
INFO - dbg state
INFO - state idle
INFO - dbg state done
INFO - start set accept filter
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - scan has 1 elements
INFO - done set accept filter
INFO - start async command
INFO - dbg state
INFO - state active
INFO - dbg state done
INFO - scan has 1 elements
INFO - scan has 1 elements
ERROR - connect: BleHost(Hci(Command Disallowed))
INFO - connecting: (AddrKind(0), BdAddr([98, 179, 117, 87, 221, 196]))
```

Easyoakland:
Update: It looks like
```
embassy_futures::yield_now().await;
embassy_futures::yield_now().await;
```
also works. I guess the pr worked because it yielded twice. Don't yet know why yielding twice works.

Easyoakland:
I'll close this pr. Not sure what to look at to debug this. If you have any suggestions let me know. Otherwise I'll use the double-yield work-around.

PR #322, opened 2025-03-20
Use user provided value directly when expanding characteristic macro

### LINKED ISSUE #320: Use value directly when expanding characteristic macro
I'm trying to create a HID service using trouble-host-macro:

```rust
const repost_desc: [u8; 90] = [...];
#[gatt_service(uuid = service::HUMAN_INTERFACE_DEVICE)]
struct HidService {
    #[characteristic(uuid = "2a4a", read, value = [0x01, 0x01, 0x00, 0x03])]
    hid_info: [u8; 4],
    #[characteristic(uuid = "2a4b", read, value = report_desc)]
    report_map: [u8; 90],
    #[characteristic(uuid = "2a4c", write_without_response)]
    hid_control_point: u8,
    #[characteristic(uuid = "2a4e", read, write_without_response, value = 1)]
    protocol_mode: u8,
    #[descriptor(uuid = "2908", read, value = [BleCompositeReportType::Keyboard as u8, 1u8])]
    #[characteristic(uuid = "2a4d", read, notify)]
    input_keyboard: [u8; 8],
    #[descriptor(uuid = "2908", read, value = [BleCompositeReportType::Keyboard as u8, 2u8])]
    #[characteristic(uuid = "2a4d", read, write, write_without_response)]
    output_keyboard: [u8; 1],
}
```
But I got the following error:

```
error[E0599]: no function or associated item named `default` found for array `[u8; 90]` in the current scope
  --> /Users/haobogu/Projects/keyboard/rmk/rmk/src/ble/trouble.rs:84:17
   |
84 |     report_map: [u8; 90],
```

I had to wrap the slice with a local struct and implement default trait for it. 

The macro expansion code is here:

https://github.com/embassy-rs/trouble/blob/84051feb1c43b5b8702a120107e715916d0422ab/host-macros/src/service.rs#L198-L207

I found it's unnecessary to create a default value then re-binding the user provided value, is there any other consideration to not use something like:

```rust
        let default_value = match characteristic.args.default_value {
            Some(val) => quote!(#val),        // if set by user
            None => quote!(<#ty>::default()), // or default otherwise
        };

        self.code_build_chars.extend(quote_spanned! {characteristic.span=>
            let (#char_name, #(#named_descriptors),*) = {
                static #name_screaming: static_cell::StaticCell<[u8; <#ty as trouble_host::types::gatt_traits::AsGatt>::MAX_SIZE]> = static_cell::StaticCell::new();
                let mut val = #default_value  // <- Use default value above directly
```

BTW, is it possible to use a `&static [u8]` as the `report_desc ` value here?

HaoboGu:
Closes #320


lulf (approved):
I'm a bit surprise it was this easy to fix. I'm curious if there is some gotcha with lifetimes and/or different types. Maybe you can check this @jamessizeland or @petekubiak

petekubiak:
Yes I'll take a look at this tomorrow 👍

petekubiak (approved):
Please change [line 200](https://github.com/embassy-rs/trouble/pull/322/files#diff-3197bd835db34bdc398d08a9cdcaf02317b03a4816e0cccadff2973fc51cf992R200) to:
```rust
None => quote_spanned!(characteristic.span => <#ty>::default()), // or default otherwise
```

This will mean that if a type is used which doesn't implement `Default` is used, the error will highlight the characteristic's type definition rather than the macro declaration.

Otherwise, looks good - great improvement!

petekubiak:
As far as I can tell this seems sound. I think the `let mut val = ...` is a holdover from before the `ToGatt` constraints were added to `Characteristic` in order to make sure that the provided value type matched the type defined for the characteristic field. Now that the characteristic holds the type info this is no longer required.

petekubiak:
Ideally I'd like to be able to inject some sort of hint to say "add the `value` argument" but type checking happens after macro expansion, and the only way I can think of to determine from within the proc macro whether a type implements a particular trait is incredibly cursed haha

HaoboGu:
Thanks for the review! I've updated it

jamessizeland:
LGTM! Yes you're right about the reason for the `let mut val = ...` and your great work has helped to remove the need!

PR #323, opened 2025-03-20
Add missing lifetime use and import

HaoboGu:
When compiling with `security` feature on latest stable(1.85.1), compiler suggests some fixes, which are included in this PR.


lulf (approved):
Thanks!

PR #329, opened 2025-03-25
Emit bonded event after bonding

### LINKED ISSUE #328: Saving bond info
I'm playing with the security feature, trying to implement a re-connectable device. 

I'm running `examples/apps/src/ble_bas_peripheral_sec.rs` by

```
cargo run --release --no-default-features --features=esp32c6,security  --target=riscv32imac-unknown-none-elf --bin ble_bas_peripheral_sec
```

but I found that I couldn't get the bond info when disconnecting:

```rust
async fn gatt_events_task<C: Controller>(
    server: &Server<'_>,
    conn: &GattConnection<'_, '_>,
    stack: &Stack<'_, C>,
) -> Result<(), Error> {
    let level = server.battery_service.level;
    loop {
        match conn.next().await {
            GattConnectionEvent::Disconnected { reason } => {
                info!("[gatt] disconnected: {:?}", reason);
                // Print the saved bond info when disconnecting
                let bond_info = stack.get_bond_information();
                info!("on disconnect Bond info: {:?}", bond_info);
                break;
            }
```

The following is the log:

```
INFO - Starting advertising and GATT service
INFO - [host] filter accept list size: 12
INFO - [host] setting txq to 24, fragmenting at 255
INFO - [host] configuring host buffers (8 packets of size 255)
INFO - [host] initialized
INFO - [host] Device Address 42:4C:CA:5B:C7:DD
TRACE - [host] enabling advertising
INFO - [adv] advertising
TRACE - [host] connection with handle ConnHandle(0) established to BdAddr([e6, cb, 86, 6d, 9b, 7c])
TRACE - [link][poll_accept] connection handle ConnHandle(0) in role Peripheral accepted
TRACE - [gatt 0] connecting to server
TRACE - [server] searching for peer BdAddr([230, 203, 134, 109, 155, 124])
INFO - [adv] connection established
INFO - [custom_task] notifying connection of tick 1
INFO - [custom_task] RSSI: 127
INFO - [host] agreed att MTU of 251
INFO - [custom_task] notifying connection of tick 2
INFO - [custom_task] RSSI: -62
INFO - [host] disconnection event on handle 0, reason: Remote User Terminated Connection
INFO - [gatt] disconnected: Err(Remote User Terminated Connection)
INFO - on disconnect Bond info: []
INFO - [gatt] task finished
INFO - Bond info: []
TRACE - [gatt 0] disconnecting from server
TRACE - [host] enabling advertising
INFO - [adv] advertising
```

I also found that the BLE device couldn't be found by my macbook and iPhone, I could only connect it using nRFConnect app.

Do I miss something to make it work?

HaoboGu:
This PR adds `Bonded` event to the connection, which would be emitted after bonding(not pairing).

`ltk` and `addr` are set public, making it easier for downstream developers to serialize `BondInformation`.

Closes #328.

### DIFF
--- a/host/src/security_manager/mod.rs
+++ b/host/src/security_manager/mod.rs
@@ -39,6 +39,8 @@ pub(crate) enum SecurityEventData {
     SendLongTermKey(ConnHandle),
     /// Enable encryption on channel
     EnableEncryption(ConnHandle),
+    /// Bonded with peer
+    Bonded(ConnHandle, BondInformation),
     /// Pairing timeout
     Timeout,
     /// Oairing timer changed
@@ -46,12 +48,12 @@ pub(crate) enum SecurityEventData {
 }
 
 /// Bond Information
-#[derive(Clone, PartialEq)]
+#[derive(Clone, Debug, PartialEq)]
 pub struct BondInformation {
     /// Long Term Key (LTK)
-    ltk: LongTermKey,
+    pub ltk: LongTermKey,
     /// Device address
-    address: BdAddr,
+    pub address: BdAddr,
     // Identity Resolving Key (IRK)?
     // Connection Signature Resolving Key (CSRK)?
 }
@@ -1079,8 +1081,9 @@ impl<const BOND_COUNT: usize> SecurityManager<BOND_COUNT> {
             (pairing_state.role, local_check)
         };
         if role == LeConnRole::Central {
-            self.store_pairing()?;
+            let bond_info = self.store_pairing()?;
             self.try_send_event(SecurityEventData::EnableEncryption(handle))?;
+            self.try_send_event(SecurityEventData::Bonded(handle, bond_info))?;
         } else {
             let mut packet = self.prepare_packet(Command::PairingDhKeyCheck, connections)?;
 
@@ -1095,7 +1098,8 @@ impl<const BOND_COUNT: usize> SecurityManager<BOND_COUNT> {
                     return Err(error);
                 }
             }
-            self.store_pairing()?;
+            let bond_info = self.store_pairing()?;
+            self.try_send_event(SecurityEventData::Bonded(handle, bond_info))?;
         }
         {
             let mut pairing_state = self.pairing_state.borrow_mut();
@@ -1149,7 +1153,7 @@ impl<const BOND_COUNT: usize> SecurityManager<BOND_COUNT> {
         Ok(())
     }
 
-    fn store_pairing(&self) -> Result<(), Error> {
+    fn store_pairing(&self) -> Result<BondInformation, Error> {
         let pairing_state = self.pairing_state.borrow();
         if let (Some(ltk), Some(peer_address)) = (pairing_state.ltk, pairing_state.peer_address) {
             let ltk = LongTermKey(ltk);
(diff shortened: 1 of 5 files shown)


host/src/security_manager/mod.rs
```diff
@@ -39,19 +39,21 @@ pub(crate) enum SecurityEventData {
     SendLongTermKey(ConnHandle),
     /// Enable encryption on channel
     EnableEncryption(ConnHandle),
+    /// Bonded with peer
+    Bonded(ConnHandle, BondInformation),
```
lulf, line 43: Instead of adding another event in the security manager for the same thing, maybe extend the EnableEncryption with the bond information?

Then, when the HCI command to enable the encryption is done (https://github.com/embassy-rs/trouble/blob/86cf58509f91e19bcbf78b7cc99bd4dcf7b84934/host/src/connection_manager.rs#L672) it can emit the Bonded event on the connection.

Otherwise you could end up in a situation where it failed to enable encryption, and still emit the bonded event.

lulf (commented):
Thanks, it looks good! I have one comment related to how the event gets created in the security manager.

host/src/security_manager/mod.rs
HaoboGu: Do you mean `EnableEncryption(ConnHandle, BondInformation)`?

I added a separate event since it's valid only when `security` feature is enabled. If `EnableEncryption(ConnHandle, BondInformation)` is used, there would be two variants of this enum variant:

```rust
    #[cfg(not(feature = "security"))]
    EnableEncryption(ConnHandle),
    #[cfg(feature = "security")]
    EnableEncryption(ConnHandle),
```

HaoboGu: In current implementation, when the role is peripheral, `EnableEncryption` event is not sent:

https://github.com/embassy-rs/trouble/blob/904171eecb04e4784d4ce8f4a9a020637634ec53/host/src/security_manager/mod.rs#L1081-L1099


Adding `EnableEncryption(ConnHandle, BondInformation)` would result in sending `EnableEncryption` event for both central and peripheral, is it correct?

lulf: Ah, I didn't read the code closely enough. Yes, checking the role like that would work I think!

lulf: Actually, I think something more is a bit strange here. The PairingState should ideally tell whether a device is bonded or not, but I don't see it ever being set to PairingState::Complete. Perhaps the appropriate place to emit this bonded event (and set the pairingstate::complete) is this code:

```
    pub(crate) fn handle_security_hci_event(&self, event: bt_hci::event::Event) -> Result<(), Error> {
        #[cfg(feature = "security")]
        {
            self.security_manager.handle_event(&event)?;

            if let bt_hci::event::Event::EncryptionChangeV1(event_data) = event {
                self.with_connected_handle(event_data.handle, |storage| {
                    storage.encrypted = event_data.enabled;
          

                    Ok(())
                })?;
            }
        }
        Ok(())
    }
``` 

The EncryptionChangeV1 is emitted the first time a connection encryption gets enabled. And it's not really until this point when receiving the event from the controller that the connection is truly encrypted.

lulf: Perhaps you have some idea what to do here @blueluna ?

blueluna: My understanding is that this shall emit an event when pairing has succeeded. I would say that the `store_pairing` function is the place where the pairing process is concluded. That would be a good place to emit an event. The `Complete` state should be removed.

The HCI encryption change event will be emitted event if there has been no pairing process, for example if there is an existing long term key in the key store.

HaoboGu: I've update the code, merged `Bonded` and `EnableEncryption`.

lulf (approved):

PR #330, opened 2025-03-27
Add function which removes bond info, update existing bond info

HaoboGu:
This PR adds `remove_bond_information` to the stack and security manager, updates `add_bond_information` to replace existing bond information when the address is same


lulf (approved):
LGTM, thank you!

PR #332, opened 2025-03-28
Disconnect when the LTK is not found

### LINKED ISSUE #331: Refuse connection when LTK is not correct
I'm trying to save multiple bonded host to BLE device, now I'm doing it my add only one bond information to the stack. For example, I have two PCs, PC1 and PC2. I'm doing the following:

1. Bond with PC1, save the bond information
2. After received switching command, the MCU would disconnect from PC1, remove bond information in the stack, then restart advertising
3. BUT, after disconnected, PC1 would also try to reconnect the device, but due to the bond information has been removed from the stack, no bond information is retrieved. So the LTK is not correct when PC1 tries to connect. I got log like this:

```rust
INFO  [adv] advertising
└─ rmk::ble::trouble::advertise::{async_fn#0} @ /Volumes/Projects/keyboard/rmk/rmk/src/ble/trouble/mod.rs:457 
INFO  [adv] connection established
└─ rmk::ble::trouble::advertise::{async_fn#0} @ /Volumes/Projects/keyboard/rmk/rmk/src/ble/trouble/mod.rs:468 
WARN  [host] Long term key request reply failed, no long term key
└─ trouble_host::connection_manager::{impl#2}::handle_security_event::{async_fn#0} @ /Users/haobogu/Projects/rust/trouble/host/src/fmt.rs:156 
```

Then the reconnection of PC1 is successful, but no data could be transmitted.

I tried to return error in `handle_security_event`:

https://github.com/embassy-rs/trouble/blob/50737e7d220e6262e6fade6f64546b44e58d0c19/host/src/connection_manager.rs#L648-L653

-> 

```
      } else {
          warn!("[host] Long term key request reply failed, no long term key")
          return Err(BleHostError::BleHost(Error::NotFound));
      }
  } else {
      warn!("[host] Long term key request reply failed, unknown peer")
  }
```

But then the whole stack panics.

Is there any way to refuse the connection in this case?

HaoboGu:
Closes #331

### DIFF
--- a/host/src/connection_manager.rs
+++ b/host/src/connection_manager.rs
@@ -622,16 +622,21 @@ impl<'d> ConnectionManager<'d> {
     ) -> Result<(), crate::BleHostError<C::Error>>
     where
         C: crate::ControllerCmdSync<bt_hci::cmd::le::LeLongTermKeyRequestReply>
-            + crate::ControllerCmdAsync<bt_hci::cmd::le::LeEnableEncryption>,
+            + crate::ControllerCmdAsync<bt_hci::cmd::le::LeEnableEncryption>
+            + crate::ControllerCmdSync<bt_hci::cmd::link_control::Disconnect>,
     {
-        use bt_hci::cmd::le::{LeEnableEncryption, LeLongTermKeyRequestReply};
+        use bt_hci::cmd::{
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks!

PR #345, opened 2025-04-07
Expose cccd table

### LINKED ISSUE #341: Save CCCD value and restore
I found the HID notify didn't work for a bonded device after I reboot the peripheral. It seems that the CCCD value of notify characteristic is wrong after the reboot. Some central implementation(windows) don't write cccd again when reconnecting a known device, so should we provide a way to persist the cccd value and restore them when reconnecting to a known central?

HaoboGu:
CCCD value should be stored on the server side during reboot, this PR exposes CCCD table which makes it possible to do that.

Closes #341


lulf (approved):
Thanks!

It feels like we should have some more generic 'load/store attribute table metadata' API, but this works for now.

HaoboGu:
> Thanks!
> 
> It feels like we should have some more generic 'load/store attribute table metadata' API, but this works for now.

Yes, I agree :D, even though the CCCD values and bond information are the main stuffs which needs to be stored across reconnections.

PR #347, opened 2025-04-08
Fix descriptor cannot be recognized by Linux/Android

HaoboGu:
Now the characteristic's descriptor has a minimal 16 bytes length, I cannot find any thing about this limitation on the [corresponding spec](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html#UUID-5262f33f-4943-5d8d-bf44-2e895938c9a1), and according to my testing, this limitation would result in `malformed att response` parsing error on Linux/Android(aka bluez) if the descriptor length is less than 16 bytes.

### DIFF
--- a/host-macros/src/service.rs
+++ b/host-macros/src/service.rs
@@ -318,9 +318,8 @@ impl ServiceBuilder {
                     quote_spanned! {characteristic.span=>
                         #identifier_assignment {
                             let value = #default_value;
-                            const CAPACITY: usize = if (#capacity) < 16 { 16 } else { #capacity }; // minimum capacity is 16 bytes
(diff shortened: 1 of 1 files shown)


lulf (approved):
Sounds good to me, CC @jamessizeland @petekubiak

petekubiak:
I remember talking to @jamessizeland about this when we were doing the implementation. I feel like there was a reason for the min length but I can't for the life of me remember what it was...

jamessizeland:
I'll take a look tomorrow and remind myself

lulf:
Ok, let's do it then I think!

PR #349, opened 2025-04-10
Use default mtu size before mtu negotiation

HaoboGu:
The default att mtu size is 23, and it will be updated after mtu negotiation. However, some central implementations don't send `ExchangeMtu` request at all(for example, android), instead they use 23 by default. In current implementation, if the trouble-host doesn't get any mtu negotiation request, it will use L2CAP_MTU - 4 by default, which might result in errors.


lulf (approved):
Nice find, thanks!

PR #351, opened 2025-04-11
fix/gatt: incorrect packet length

drindr:
Packet length is incorrect, when the size of ATT value is larger than MTU. The BLE stack's parser of Android rely on the packet length field strictly. With incorrect value, it will trigger error and fail to go ahead.


lulf (approved):
Nice catch, thanks!

PR #364, opened 2025-04-21
feat: allow updating advertisement data while advertising.

augustoroman:
This adds a new `update_adv_data` method to Peripheral that allows updating advertisement data without otherwise affecting any active advertisements.

### DIFF
--- a/host/src/peripheral.rs
+++ b/host/src/peripheral.rs
@@ -247,6 +247,46 @@ impl<'d, C: Controller> Peripheral<'d, C> {
         })
     }
 
+    /// Update the extended advertisment adv_data and/or scan_data for multiple
+    /// advertising sets. Does not change any other advertising parameters. If
+    /// no advertising is active, this will not produce any observable effect.
+    /// This is typically useful when implementing a BLE beacon that only
+    /// broadcasts advertisement data and does not accept any connections.
+    pub async fn update_adv_data_ext<'k>(
+        &mut self,
+        sets: &[AdvertisementSet<'k>],
+        handles: &mut [AdvSet],
+    ) -> Result<(), BleHostError<C::Error>>
+    where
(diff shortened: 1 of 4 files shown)


host/src/peripheral.rs
```diff
@@ -106,6 +106,38 @@ impl<'d, C: Controller> Peripheral<'d, C> {
...
+    /// other advertising parameters. If no advertising is active, this will not
+    /// produce any observable effect. This is typically useful when
+    /// implementing a BLE beacon that only broadcasts advertisement data and
+    /// does not accept any connections.
+    pub async fn update_adv_data<'k>(&mut self, data: Advertisement<'k>) -> Result<(), BleHostError<C::Error>>
+    where
+        C: for<'t> ControllerCmdSync<LeSetAdvData>
+            + ControllerCmdSync<LeSetAdvParams>
+            + for<'t> ControllerCmdSync<LeSetAdvEnable>
+            + for<'t> ControllerCmdSync<LeSetScanResponseData>,
```
lulf, line 119: I don't see any reason why you need to have constraints other than LeSetScanResponseData and LeSetAdvData here.

augustoroman:
Updated. I removed the extraneous constraints, added `update_adv_data_ext`, and I also added a beacon example.

augustoroman:
Ooops, missed that. Fixed -- now I have a green check.

augustoroman:
Huh. I can't see the output of the failed test. What went wrong? Can I run this locally?

lulf (approved):
The CI run can be found here https://github.com/embassy-rs/trouble/actions but it lacks sufficient data to locate, so you have to guess which one it is. But in any case, the CI issue was because I had an example running locally and forgot to unplug, which distracted the CI examples from working :joy:

PR #365, opened 2025-04-22
feat: allow configuring custom pool allocator

lulf:
* Add Packet trait which constrains the packet allocator.
* Use Packet trait in all places needed.
* Provide out of the box configuration-based packet pool.
* Adjust host buffer settings

This initial work provides a default packet pool that is used by examples and tuned according to the current pool.  A custom allocator example is shown in the serial-hci high throughput examples.


lulf:
This 'infects' most of the types to carry the additional 'PacketPool' generic. Downside is this makes it a bit more verbose. Upside is that you can plug a custom allocator, and paves the way for a big optimization where segmentation/reassembly is done in the host runner, allowing accepting/passing the processed packets to/from the application and combine with other parts of a system.

PR #367, opened 2025-04-22
Add IRK support

HaoboGu:
- basic IRK crypto stuffs
- Resolve RPA using IRK
- Fix `ConnectionEventData` typo

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -70,7 +70,7 @@ jobs:
           RUST_TEST_THREADS: 1
         run: |
           cd host
-          cargo test --features log,security --test '*' -- --nocapture
+          cargo test --features log --test '*' -- --nocapture
       - name: Update failed status
         if: failure()
         env:


HaoboGu:
I think I've finished the calculation part, but I'm having problems when reconnecting, the att server always reports full-slot:

https://github.com/embassy-rs/trouble/blob/563f28f6d194653b082d40fd8511f04e1942d904/host/src/attribute_server.rs#L117-L133

It seems that the CccdTables checks the addr directly. With IRK, the addr should be checked using IRK key if it's RPA. @lulf Do you have any suggestions to inject IRK to `CccdTables`?

lulf:
> IRK key

Maybe instead of BdAddr, create some kind of DeviceIdentity enum that can be either the address or the IRK?

HaoboGu:
Let me try!

HaoboGu:
I've added the `Identity` replacing previouse `BdAddr`.

After I merged the main branch, it reports many errors about the `<P>`, I guess the security is not updated by #365

HaoboGu:
I think it works now

lulf (approved):

lulf:
```
thread 'gatt_client_server' panicked at /home/ghrunner/_work/trouble/trouble/host/src/lib.rs:565:17:
The security manager random number generator has not been seeded from a cryptographically secure random number generator
[2025-04-23T18:44:22Z TRACE mio::poll] deregistering event source from poller

thread 'gatt_client_server' panicked at /home/ghrunner/_work/trouble/trouble/host/src/lib.rs:565:17:
The security manager random number generator has not been seeded from a cryptographically secure random number generator
[2025-04-23T18:44:22Z TRACE mio::poll] deregistering event source from poller
```

HaoboGu:
🤯 I didn't touch the random seed part, @lulf could you give me some hints about this error? Thanks!

HaoboGu:
Oh, I added `security` feature to the test.. Removed it.

HaoboGu:
weird, I've removed the "security " feature in ci.sh, but it still reports the random seed is not set error

lulf:
This is because it's merged to master and it uses the tests.yaml from there when running the tests. We can just merge it.

HaoboGu:
thanks!

PR #368, opened 2025-04-23
Fix packet pool error in security

HaoboGu:


HaoboGu:
I have to say, on the user side, lots lots of generic parameters are added, which is painful to correct all of them.. 🥲

lulf:
Yes, its unfortunate. But i didnt really see a way around it. Sorry about that! Maybe some default generic could work.

PR #369, opened 2025-04-23
fix: Don't panic when parsing bad advertisement data

augustoroman:
When parsing advertisements, it's possible that the advertisement packet is not correctly encoded and the decode function will be called when there is insufficient data to decode.

These cases should return an error instead of panicking because of an OOB memory access.


lulf (approved):
Thanks for the fix!

lulf:
The CI only obeys those with write access to the repo, as it's running on an github runner in my home :)

PR #376, opened 2025-04-28
Fix RX ACL disconnection behavior

timokroeger:
Dropping a connection during ongoing ATT data transmission causes a panic. The connection drop code sends a disconnect command to the BT controller. We still continue to receive buffered ACL packets from the BT controller.
The code deleted in this commit triggers a second disconnect command which runs into a panic.
Disconnects are correctly handled by the DisconnectionComplete event. There should be no need to keep the special disconnect logic in the ACL RX path.

EDIT: The PR now includes an additional commit.
The wake to actually send the disconnect command to the BLE controller was missing.

### DIFF
--- a/host/src/host.rs
+++ b/host/src/host.rs
@@ -748,13 +748,9 @@ impl<'d, C: Controller, P: PacketPool> RxRunner<'d, C, P> {
 
                         if let Error::Disconnected = e {
                             warn!("[host] requesting {:?} to be disconnected", acl.handle());
-                            let _ = host
-                                .command(Disconnect::new(
-                                    acl.handle(),
-                                    DisconnectReason::RemoteUserTerminatedConn,
-                                ))
-                                .await;
                             host.connections.log_status(true);
+                            host.connections
+                                .request_handle_disconnect(acl.handle(), DisconnectReason::RemoteUserTerminatedConn);
                         }
 
                         let mut m = host.metrics.borrow_mut();


timokroeger:
The panic was actually from my code asserting that `run()` should never return.
This is my log:
```
9.100219 INFO  [ble] button pressed, disconnecting
└─ ioutble::tasks::ble::button_pressed::{async_fn#0} @ /home/user/code/ioutble/firmware\src/tasks\ble.rs:246
9.100280 TRACE [gatt 1] disconnecting from server
└─ trouble_host::gatt::{impl#0}::drop @ /home/user/code/trouble/host\src\fmt.rs:117
9.100341 INFO  [ble] disconnected
└─ ioutble::tasks::ble::ble::{async_fn#0}::{async_block#0} @ /home/user/code/ioutble/firmware\src/tasks\ble.rs:163
9.100372 DEBUG [ble] waiting for disconnection to finish
└─ ioutble::tasks::ble::ble::{async_fn#0}::{async_block#0} @ /home/user/code/ioutble/firmware\src/tasks\ble.rs:170
9.106719 WARN  [host] encountered error processing ACL data for ConnHandle(1): Disconnected
└─ trouble_host::host::{impl#4}::run_with_handler::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:156
9.106781 WARN  [host] requesting ConnHandle(1) to be disconnected
└─ trouble_host::host::{impl#4}::run_with_handler::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:156
9.106903 DEBUG [link][idx = 0] state = state = DisconnectRequest(RemoteUserTerminatedConn), conn = Some(ConnHandle(1)), flow = 3, role = Some(Peripheral), peer = Some(BdAddr(BdAddr([58, 72, 9A, 70, 3B, 2C])) ), ref = 0
└─ trouble_host::connection_manager::{impl#0}::print @ /home/user/code/trouble/host\src\fmt.rs:130
9.107269 TRACE [host] poll disconnecting links
└─ trouble_host::host::{impl#5}::run::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:117
9.107421 TRACE [host] control_fut exit
└─ trouble_host::host::{impl#3}::run_with_handler::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:117
9.120635 ERROR panicked at 'BLE task ended unexpectedly'
└─ ioutble::__ble_task_task::{async_fn#0} @ /home/user/code/ioutble/firmware\src\main.rs:118
```

The runner exits because I receive `BleHost(Hci(Command Disallowed))` here:
https://github.com/embassy-rs/trouble/blob/8062307ff99c25d43654943442a37c4a50785090/host/src/host.rs#L1036-L1043

BT Controller in question is Softdevice on a nrf52 chip.

Interestingly enough the problem appears to be fixed with just the second commit which triggers the disconnect waker.

lulf:
> The panic was actually from my code asserting that `run()` should never return. This is my log:
> 
> ```
> 9.100219 INFO  [ble] button pressed, disconnecting
> └─ ioutble::tasks::ble::button_pressed::{async_fn#0} @ /home/user/code/ioutble/firmware\src/tasks\ble.rs:246
> 9.100280 TRACE [gatt 1] disconnecting from server
> └─ trouble_host::gatt::{impl#0}::drop @ /home/user/code/trouble/host\src\fmt.rs:117
> 9.100341 INFO  [ble] disconnected
> └─ ioutble::tasks::ble::ble::{async_fn#0}::{async_block#0} @ /home/user/code/ioutble/firmware\src/tasks\ble.rs:163
> 9.100372 DEBUG [ble] waiting for disconnection to finish
> └─ ioutble::tasks::ble::ble::{async_fn#0}::{async_block#0} @ /home/user/code/ioutble/firmware\src/tasks\ble.rs:170
> 9.106719 WARN  [host] encountered error processing ACL data for ConnHandle(1): Disconnected
> └─ trouble_host::host::{impl#4}::run_with_handler::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:156
> 9.106781 WARN  [host] requesting ConnHandle(1) to be disconnected
> └─ trouble_host::host::{impl#4}::run_with_handler::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:156
> 9.106903 DEBUG [link][idx = 0] state = state = DisconnectRequest(RemoteUserTerminatedConn), conn = Some(ConnHandle(1)), flow = 3, role = Some(Peripheral), peer = Some(BdAddr(BdAddr([58, 72, 9A, 70, 3B, 2C])) ), ref = 0
> └─ trouble_host::connection_manager::{impl#0}::print @ /home/user/code/trouble/host\src\fmt.rs:130
> 9.107269 TRACE [host] poll disconnecting links
> └─ trouble_host::host::{impl#5}::run::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:117
> 9.107421 TRACE [host] control_fut exit
> └─ trouble_host::host::{impl#3}::run_with_handler::{async_fn#0} @ /home/user/code/trouble/host\src\fmt.rs:117
> 9.120635 ERROR panicked at 'BLE task ended unexpectedly'
> └─ ioutble::__ble_task_task::{async_fn#0} @ /home/user/code/ioutble/firmware\src\main.rs:118
> ```
> 
> The runner exits because I receive `BleHost(Hci(Command Disallowed))` here:
> 
> https://github.com/embassy-rs/trouble/blob/8062307ff99c25d43654943442a37c4a50785090/host/src/host.rs#L1036-L1043
> 
> BT Controller in question is Softdevice on a nrf52 chip.
> 
> Interestingly enough the problem appears to be fixed with just the second commit which triggers the disconnect waker.

Thanks, that makes sense... it's probably closing the race condition to the minimal. 

I'm a bit puzzled why there's a Command Disallowed and not a 'Unknown connection identifier' error in this case. If trying to disconnect an unknown handle it returns 'unknown connection identifier' (which is why this is now handled in the control_fut disconnecting case, but if it tries to disconnect a handle twice, it returns 'command disallowed'.

I'm wondering if we should instead continue in the control_fut if receiving a 'command disallowed' rather than returning the error. 

Having seen issues with inconsistent state in the past (though this was way back, maybe it's no longer an issue), I'm a bit worried about removing it.

Thoughts?

timokroeger:
Retrying with the waker fix (order of disconnect commands changes):
1. When dropping a connection handle we now disconnect before receiving any further ACL data (discommand command success)
2. On ACL rx we detect that the data belongs to disconnected channel
3. We send another disconnect command (to ensure sync of host and contoller). This command also gets a `BleHost(Hci(Command Disallowed))` but its result is discarded.

I agree that the expected response should be `Unknown connection identifier` rather than `Command disallowed` but thats we get unfortunately.

> I'm wondering if we should instead continue in the control_fut if receiving a 'command disallowed' rather than returning the error.

We should not observe the `command disallowed` anymore after the waker fix when we can assume that the control runner is polled before the rx runner. I don’t think we need to change the behavior.

Its probably better to drop the first commit and open a new PR with just the waker fix. What do you think?

lulf:
Actually, thinking about this some more. What we really want is to only run this if there is no disconnection in progress. One way would be to add another Error variant for Disconnecting, but I think that's not a great UX from the application side which usually doesn't care about the difference. 

So maybe adding an is_disconnecting(handle: ConnHandle) to the connection_manager, which the rx fut can check when it gets the disconnect error. If is_disconnection() == true, then it means that the control_fut will issue the disconnection anyway.

timokroeger:
Had to force-push once more to fix the commit message, should be ready to review again

Issue #371, opened 2025-04-24
Invalid PSM used in L2CAP example


bittailor:
Hi 

I think the current ble L2CAP examples use an _invalid_ `PSM` value. 

https://github.com/embassy-rs/trouble/blob/2128a22360300554b90efaa7804c759b24a60fad/examples/apps/src/ble_l2cap_peripheral.rs#L54

It works when both, central and peripheral, are done with trouble on an embedded device. But when I try to connect from macOS as central (using CoreBluetooth in a Swift test client) to a peripheral done with the trouble example code, the CoreBluetooth call 

``` swift
peripheral.openL2CAPChannel(0x2349)
```

fails with

```
Failed to open L2CAP channel Optional(Error Domain=CBErrorDomain Code=1 "One or more parameters were invalid." UserInfo={NSLocalizedDescription=One or more parameters were invalid.})
```

When looking at the [Bluetooth Specification for the PSM (or rather SPSM as it's called for LE L2CAP)](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-60/out/en/host/logical-link-control-and-adaptation-protocol-specification.html#UUID-1ffdf913-7b8a-c7ba-531e-2a9c6f6da8fb) it should be in the range of `0x0000-0x00FF` or rather `0x0080-0x00FF` for one not assigned by SIG.

![Image](https://github.com/user-attachments/assets/2989e25d-0c35-47df-ba1f-87be2212616f)

So when I change the PSM from `0x2349` to `0x81` in the trouble example 

``` rust
let mut ch1 = L2capChannel::accept(&stack, &conn, &[0x81], &Default::default())
                .await
                .unwrap();
```

and also in my Swift test client, the CoreBluetooth call

``` swift
peripheral.openL2CAPChannel(0x81)
```
succeeds and I can exchange data.

lulf:
Huh, I didn't know about that restriction. Can you open a PR to fix it?

bittailor:
> Huh, I didn't know about that restriction. Can you open a PR to fix it?

👍 will do so.

bittailor:
#377

PR #377, opened 2025-04-28
Use Bluetooth specification conform L2CAP PSM

bittailor:
Use a PSM from the dynamic range (0x0080-0x00FF) according to the [Bluetooth Specification for L2CAP channels using LE Credit Based Flow Control mode](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-60/out/en/host/logical-link-control-and-adaptation-protocol-specification.html#UUID-1ffdf913-7b8a-c7ba-531e-2a9c6f6da8fb).

With this the L2CAP peripheral examples also work with _non embedded_ centrals, like a Swift test client running on macOS using [CBPeripheral/openL2CAPChannel ](https://developer.apple.com/documentation/corebluetooth/cbperipheral/openl2capchannel(_:))

See #371

### DIFF
--- a/examples/apps/src/ble_l2cap_central.rs
+++ b/examples/apps/src/ble_l2cap_central.rs
@@ -2,6 +2,8 @@ use embassy_futures::join::join;
 use embassy_time::{Duration, Timer};
 use trouble_host::prelude::*;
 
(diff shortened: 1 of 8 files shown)


lulf (approved):
Thank you!

PR #382, opened 2025-04-30
Replace unmaintained inflector dependency with convert_case

ratmice:
I wasn't sure how to test this change, as `cargo test` in the host-macros directory didn't actually succeed before or after this.

The repository for the inflector crate at https://github.com/whatisinternet/Inflector
was archived last october, and hasn't been updated for 7 years.

This changes it to use the https://crates.io/crates/convert_case which was the most maintained crate I could find offering
the functionality, it is used as an optional dependency by the `derive-more` crate.

The relevant documentation for the before/after.
https://docs.rs/Inflector/latest/inflector/cases/screamingsnakecase/fn.to_screaming_snake_case.html
https://docs.rs/convert_case/latest/convert_case/enum.Case.html#variant.Constant

I'm assuming the usage of the alloc crate in the proc-macro is okay since it seemed like inflector also relied upon std.

### DIFF
--- a/host-macros/Cargo.toml
+++ b/host-macros/Cargo.toml
@@ -8,11 +8,11 @@ keywords = ["no-std"]
 categories = ["embedded", "hardware-support", "no-std"]
 
 [dependencies]
+convert_case = "0.8.0"
 syn = { version = "^2", features = ["full", "extra-traits"] }
 quote = "1.0.7"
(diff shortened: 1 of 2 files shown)


lulf (approved):

lulf:
Great, thanks for fixing!

PR #385, opened 2025-05-04
fix: use length from AdStructure::encode_slice in examples/tests

eldad:
When the length is not used, the stack will emit a packet with padding, which wastes bandwidth. Another risk that can arise if the example code is followed elsewhere, is when buffers are reused as `encode_slice` does not zero out the rest of the buffer.


lulf (approved):
Thanks!

PR #392, opened 2025-05-15
Make the server field of the GATT Server struct public so it can be used across modules

### LINKED ISSUE #391: #[gatt_server] macro generates a private AttributeServer field, limiting usability across modules
Presently, the `#[gatt_server]` macro always generates a `server` field (of type `AttributeServer<...>`) as a _private_ field inside the resulting struct.

This becomes problematic when the generated struct is defined in its own module and needs to be passed to APIs like `.with_attribute_server(...)`, which require direct access to the internal `AttributeServer`. For example:

```
66 |     .with_attribute_server(&server.server)?;
   |                                    ^^^^^^ private field

```

Attempting to pass `&server` directly results in a type mismatch, which can be even more obscure:

```
error[E0308]: mismatched types
    |
63  |     let conn = advertiser.accept().await?.with_attribute_server(&server)?;
    |                                           --------------------- ^^^^^^^ expected `&AttributeServer<'_, _, P, _, _, _>`, found `&Server<'_>`
    |                                           |
    |                                           arguments to this method are incorrect
    |
    = note: expected reference `&AttributeServer<'_, _, P, _, _, _>`
               found reference `&server::Server<'_>`

    |
333 |     pub fn with_attribute_server<
    |            ^^^^^^^^^^^^^^^^^^^^^
```

This effectively forces the user to define the GATT server struct in the same module where it’s being used — a restriction that limits code organization.

Perhaps we should make the `server` field `pub`, or add a macro parameter to control field visibility, such as `server_visibility = "pub"`?

This would allow the generated struct to be cleanly defined in a separate module and used elsewhere without leaking implementation details or resorting to awkward workarounds.

I'm happy to contribute the change as long as we're aligned on how we wanna approach this.

petekubiak:
Resolves #391

### DIFF
--- a/host-macros/src/server.rs
+++ b/host-macros/src/server.rs
@@ -158,7 +158,7 @@ impl ServerBuilder {
 
             #visibility struct #name<'values>
             {
-                server: trouble_host::prelude::AttributeServer<'values, #mutex_type, #packet_type, _ATTRIBUTE_TABLE_SIZE, _CCCD_TABLE_SIZE, _CONNECTIONS_MAX>,
+                pub server: trouble_host::prelude::AttributeServer<'values, #mutex_type, #packet_type, _ATTRIBUTE_TABLE_SIZE, _CCCD_TABLE_SIZE, _CONNECTIONS_MAX>,
                 #code_service_definition
             }
 


sssilver (approved):

sssilver:
Amazing, thank you so much

PR #394, opened 2025-05-19
Tidy up ble bas peripheral example

jamessizeland:
Small refactor of the gatt peripheral examples to tidy up their structure.  I also tend to implement advertiser as a method on the Server, so wanted your thoughts on if this is a patter you're happy with.  (I think we might also move to having a macro generated advertiser function in future too...)

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -52,7 +52,7 @@ where
 
     let _ = join(ble_task(runner), async {
         loop {
-            match advertise("Trouble Example", &mut peripheral, &server).await {
+            match server.advertise("Trouble Example", &mut peripheral).await {
                 Ok(conn) => {
                     // set up tasks when the connection is established to a central, so they don't run when no one is connected.
                     let a = gatt_events_task(&server, &conn);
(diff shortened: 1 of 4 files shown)


lulf:
Nice to get it tidied up, thanks! I disagree with having advertise() on server. I think it's mixing GATT and advertising too much, and they're really separate things, although the advertisement payload can 'hint' to what the server supports. I think it's better to keep at least advertise similar across examples that use/do not use GATT.

jamessizeland:
Reverted. The main thing we need to add to the advertise function is a way to programmatically update the AdStructure::ServiceUuids16 to include the Uuids of the services on our server.  That could be made up as a method on Server in the macro i.e.

```rust
let len = AdStructure::encode_slice(
        &[
            AdStructure::ServiceUuids16(&server.list_services()),
            AdStructure::CompleteLocalName(name.as_bytes()),
        ],
        &mut advertiser_data[..],
    )?;
```

jamessizeland:
I'm not sure what this CI build error is, looks like a system error rather than the code.

```
[build](https://github.com/embassy-rs/trouble/actions/runs/15134106916/job/42542482261)
System.IO.IOException: No space left on device
```

lulf:
I cleaned the cache and reran the job, seems like there is a different error related to the toolchain and clippy.

PR #404, opened 2025-05-30
Simplify and generify GattEvent; allow again for custom processing of GATT events

ivmarkov:
[`GattEvent`](https://github.com/embassy-rs/trouble/blob/cb4b2cd424cdb2af54bad52922798a58d8786d7f/host/src/gatt.rs#L234) is currently an enum with only two variants:
- `Write` - captures ATT Write requests and Write commands
- `Read` - captures ATT Read and ReadBlob requesats

(1) This modeling is _incomplete_. 

There are many more ATT payloads that are currently impossible to capture with `GattEvent`. Rather than extending `GattEvent` to model all of those, the GATT layer currently just doesn't model these additional ATT payloads with a `GattEvent` _at all_ and instead [directly processes those payloads](https://github.com/embassy-rs/trouble/blob/cb4b2cd424cdb2af54bad52922798a58d8786d7f/host/src/gatt.rs#L222) _without even notifying the user that such a processing is due / had happened_.

(2) Another problem with `GattEvent` is that if the user code would like to - for whatever reason - not call `GattEvent::accept` or `GattEvent::reject` but do its own processing of the original `PDU` / `GattData` payload, it is currently just not possible, as once a `GattEvent` is constructed by `GattConnection`, it is either `accept` or `reject`, no other way.

(1) and (2) cause problems to `rs-matter` which wants to do its own custom processing for a handful of ATT payloads, while still delegating the majority of the ATT payloads to a proc-macro derived attribute server which saves a great deal of boilerplate code for your typical meta-data "what attributes are in there?" type of ATT queries.

===

This PR is solving both (1) and (2) as follows:

For (1), the PR
- Converts `GattEvent` from an enum into a struct with two fields: `payload: GattData` (the GATT payload) and `server` - the type-erased attribute server
- **Removes** the silent, direct processing of "other" payloads from `GattConnection::next`, which I think is a big win. ALL ATT payloads are now converted to a `GattEvent` and are only processed if the user calls `accept` (or on drop, as before)
- `ReadEvent` and `WriteEvent` are now gone. They were anyway pre-dating the `AttClient`/`AttServer` enumerations which were introduced with PR #267 , and with that PR in place, they seem obsolete
- Rather than `match`-ing on `GattEvent::Read` and `GattEvent::Write`, user can now match on `GattEvent::payload().incoming()` which does provide a **complete** modeling of the ATT payload, and not just on "read" and "write", where the fidelity of the exact read and write ATT operation was even lost

For (2), the PR
- Introduces a new method, `GattEvent::into_payload` which converts the `GattEvent` instance back into the `GattData` payload it encloses, thus leaving it up to the caller to handle the payload however she sees fit. Given that #267 also introduced `GattData::reply` and `GattData::send`, this is completely possible.

===

Pros and cons:
(+) No more "silent processing bypassing GattEvent"
(+) For accepting/rejecting/`into_payload`-ing GATT events, user can now inspect/query the payload of GattEvent at a much greater detail, by matching on all possible ATT payloads supported by BLE
(+) Possibility to to do custom processing of the GATT event by `GattEvent::into_payload`
(-) Inspecting the payload of the `GattEvent` is more verbose now, as can be seen in the unit tests and examples, as e.g. "reads" and "writes" in the ATT layer can come in different shapes (i.e. a "write" could be a write request or a write command; a "read" could be a read request, or a read blob request, and so on)

### DIFF
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -1,6 +1,7 @@
 use embassy_futures::join::join;
 use embassy_futures::select::select;
 use embassy_time::Timer;
+use trouble_host::att::{AttClient, AttCmd, AttReq};
 use trouble_host::prelude::*;
 
 /// Max number of connections
@@ -106,20 +107,22 @@ async fn gatt_events_task<P: PacketPool>(server: &Server<'_>, conn: &GattConnect
     let reason = loop {
         match conn.next().await {
             GattConnectionEvent::Disconnected { reason } => break reason,
-            GattConnectionEvent::Gatt { event: Err(e) } => warn!("[gatt] error processing event: {:?}", e),
-            GattConnectionEvent::Gatt { event: Ok(event) } => {
-                match &event {
-                    GattEvent::Read(event) => {
-                        if event.handle() == level.handle {
+            GattConnectionEvent::Gatt { event } => {
+                match event.payload().incoming() {
+                    AttClient::Request(AttReq::Read { handle, .. })
+                    | AttClient::Request(AttReq::ReadBlob { handle, .. }) => {
+                        if handle == level.handle {
                             let value = server.get(&level);
-                            info!("[gatt] Read Event to Level Characteristic: {:?}", value);
+                            info!("[gatt] Read from Level Characteristic: {:?}", value);
                         }
                     }
-                    GattEvent::Write(event) => {
-                        if event.handle() == level.handle {
-                            info!("[gatt] Write Event to Level Characteristic: {:?}", event.data());
+                    AttClient::Request(AttReq::Write { handle, data })
+                    | AttClient::Command(AttCmd::Write { handle, data }) => {
+                        if handle == level.handle {
+                            info!("[gatt] Write to Level Characteristic: {:?}", data);
                         }
                     }
+                    _ => {}
                 };
                 // This step is also performed at drop(), but writing it explicitly is necessary
                 // in order to ensure reply is sent.
--- a/examples/apps/src/ble_bas_peripheral.rs
+++ b/examples/apps/src/ble_bas_peripheral.rs
@@ -1,7 +1,6 @@
 use embassy_futures::join::join;
 use embassy_futures::select::select;
 use embassy_time::Timer;
-use trouble_host::att::{AttClient, AttCmd, AttReq};
 use trouble_host::prelude::*;
 
 /// Max number of connections
@@ -108,18 +107,16 @@ async fn gatt_events_task<P: PacketPool>(server: &Server<'_>, conn: &GattConnect
         match conn.next().await {
             GattConnectionEvent::Disconnected { reason } => break reason,
             GattConnectionEvent::Gatt { event } => {
-                match event.payload().incoming() {
-                    AttClient::Request(AttReq::Read { handle, .. })
-                    | AttClient::Request(AttReq::ReadBlob { handle, .. }) => {
-                        if handle == level.handle {
+                match &event {
+                    GattEvent::Read(event) => {
+                        if event.handle() == level.handle {
                             let value = server.get(&level);
-                            info!("[gatt] Read from Level Characteristic: {:?}", value);
+                            info!("[gatt] Read Event to Level Characteristic: {:?}", value);
                         }
                     }
-                    AttClient::Request(AttReq::Write { handle, data })
-                    | AttClient::Command(AttCmd::Write { handle, data }) => {
-                        if handle == level.handle {
-                            info!("[gatt] Write to Level Characteristic: {:?}", data);
+                    GattEvent::Write(event) => {
+                        if event.handle() == level.handle {
+                            info!("[gatt] Write Event to Level Characteristic: {:?}", event.data());
                         }
                     }
                     _ => {}
--- a/examples/apps/src/ble_bas_peripheral_sec.rs
+++ b/examples/apps/src/ble_bas_peripheral_sec.rs
@@ -2,7 +2,6 @@ use embassy_futures::join::join;
 use embassy_futures::select::select;
 use embassy_time::Timer;
 use rand_core::{CryptoRng, RngCore};
-use trouble_host::att::{AttClient, AttCmd, AttReq};
 use trouble_host::prelude::*;
 
 /// Max number of connections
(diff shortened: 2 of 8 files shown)


examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -106,20 +107,22 @@ async fn gatt_events_task<P: PacketPool>(server: &Server<'_>, conn: &GattConnect
     let reason = loop {
         match conn.next().await {
             GattConnectionEvent::Disconnected { reason } => break reason,
-            GattConnectionEvent::Gatt { event: Err(e) } => warn!("[gatt] error processing event: {:?}", e),
-            GattConnectionEvent::Gatt { event: Ok(event) } => {
-                match &event {
-                    GattEvent::Read(event) => {
-                        if event.handle() == level.handle {
+            GattConnectionEvent::Gatt { event } => {
+                match event.payload().incoming() {
```
lulf: This is a little bit footgun of an API in my mind. For a a GattConnection, the payload will always be incoming, so it seems wrong to allow choosing to do anything else than treat it as a AttClient::Request.

examples/apps/src/ble_bas_peripheral.rs
```diff
@@ -106,20 +107,22 @@ async fn gatt_events_task<P: PacketPool>(server: &Server<'_>, conn: &GattConnect
...
-                    GattEvent::Write(event) => {
-                        if event.handle() == level.handle {
-                            info!("[gatt] Write Event to Level Characteristic: {:?}", event.data());
+                    AttClient::Request(AttReq::Write { handle, data })
+                    | AttClient::Command(AttCmd::Write { handle, data }) => {
+                        if handle == level.handle {
+                            info!("[gatt] Write to Level Characteristic: {:?}", data);
                         }
                     }
+                    _ => {}
```
lulf: For most users, won't this API be a lot more verbose than strictly necessary? I keep thinking it would be nice to at least have some higher level Read/Write categorization like the original nrf-softdevice had.

lulf (commented):
I like the simplifications, but there are two things I don't like that happened to the API that I hope it's possible to improve while keeping the simplifications.

examples/apps/src/ble_bas_peripheral.rs
ivmarkov: I'm not sure I understand as to where the problem is?

For a `GattConnection`, the payload will always be incoming sure, and there is anyway no "outgoing" method on `GattData` if that's the issue.

If the issue is that the user can match not just on `AttClient::Request` but also on `AttClient::Command` or `AttClient::Confirmation`, this is IMO exactly as it should be, because incoming (client -> server) ATT payloads are _not_ just "requests". You might receive commands too, as well as confirmations. In fact, `rs-matter`'s custom logic matches on `AttClient::Confirmation`.

examples/apps/src/ble_bas_peripheral.rs
ivmarkov: ... i.e. restoring the `ReadEvent` and `WriteEvent`. OK but what do we do with the "other" events which are neither "single attribute read/read blob" (`ReadEvent`) nor "single attribute write request/write command" (`WriteEvent`)? For example:
- Read by group or read by type requests
- Find by type requests
- Confirmations
- FindInfo
- etc. etc.

The whole change suggested here is around the idea that **all** ATT payloads are passed to the user in the form of a `GattEvent` so that the user has the right to accept/reject/custom-process **all**/any ATT payload.

Can you point a link to the original `nrf-softdevice` code so that I can check what they were doing?

ivmarkov: Note also that the old `ReadEvent`/`WriteEvent` logic was also destructuring the `GattData` in an irreversible way. Which made impossible for the user to do any custom processing on the incoming data. It was either accept or reject and that's it, but there was no way to "bypass" the attribute server (the new method `GattEvent::into_payload() -> GattData`), and then call `GattData::reply` or `GattData::send_unsolicited` because `GattData` itself was lost in the process of converting it to `ReadEvent` / `WriteEvent` / silently processing all others with the attribute server.

UPDATE: Actually I just checked again and the PDU was preserved in `GattEvent::Read(ReadEvent)` and in `GattEvent::Write(WriteEvent)` so we can reconstruct `GattData` from these if we want to. But the fact that the previous representation of GattEvent was not a faithful representation of all possible ATT payloads and most payloads were silently processed rather than handed off to the user in the form of a GattEvent remains.

examples/apps/src/ble_bas_peripheral.rs
ivmarkov: Perhaps [this comment of mine from the older PR](https://github.com/embassy-rs/trouble/pull/267#issuecomment-2629085279) might help too? client -> server ATT payloads are not just requests. That's why what used to be called `AttReq` previously, was renamed by that older PR to `AttClient`. As it is really not only requests that might arrive in the peripheral.

ivmarkov:
@lulf Copying here from the Matrix channel in an attempt to get feedback:

After all we discussed on the channel, perhaps two options:
(1) Enhance GattEvent with GattEvent:::Other(OtherEvent) variant
(2) Make some types / fn members public again so that I can arrange my own, alternative processing of PDUs

While (2) is generally less work, I have a slight preference for (1) as in my eyes with (1) we'll end up with a simpler, easier to explain GattEvent metaphor. There would be no more "yeah, but some events are not even "seen" by the user, and she can't veto/custom process these" hack.

What would you prefer?

Background: https://matrix.to/#/!YoLPkieCYHGzdjUhOK:matrix.org/$LzUO9INjJTFVkEue1Hd45FZwJMraUY47kRhWm61JpYE?via=matrix.org&via=beeper.com&via=tchncs.de

lulf:
I have a preference for option 1, thanks! (And sorry again for the delay)

examples/apps/src/ble_bas_peripheral.rs
ivmarkov: Topic no longer applicable after the switch to the `GattEvent::Other` approach.

examples/apps/src/ble_bas_peripheral.rs
ivmarkov: Ditto - topic no longer applicable after the switch to the `GattEvent::Other` approach.

ivmarkov:
@lulf This is ready for review again.

(I've kept the `OtherEvent` approach as a separate - 2nd - commit, so that you can compare with the original approach if you'd like. If we decide for merging, we need to squash the two.)

I think the "OtherEvent" thing turned out not-too-bad, even when comparing with the first approach. Other than introducing this additional enum variant, the following changes:
- `async fn GattData::process` is gone, as nothing is pre-processed anymore. Simply a `GattData` instance is wrapped as a `GattEvent` (hence the new `GattEvent::new` fn); all processing now is either with `GattEvent::accept`, `GattEvent::reject`, on-drop, or if the user decided to do her own thing with `GattEvent::into_payload`
- Private fns `process`, `process_accept` and `process_reject` simplified a bit in terms of their signatures
-  `ReadEvent::handle` and `WriteEvent::handle` are no more, replaced with a `handle()` method which retrieves the handle from the PDU directly. Ditto for `GattEvent::handle` which of course returns `Option<u16>` as not all payloads do have a handle

lulf (approved):
I like this approach as well, thank you for the doing the work! 🚀

PR #411, opened 2025-06-18
Use l2cap signaling when LeConnUpdate is not available

### LINKED ISSUE #338: `BleHost(Disconnected)` when setting connection param
I'm testing nrf-sdc updating connection parameter after GattConnection is established. I use
```rust
conn.raw().update_connection_params(&stack,
                &ConnectParams {
                    min_connection_interval: Duration::from_micros(7500),
                    max_connection_interval: Duration::from_micros(7500),
                    max_latency: 99,
                    event_length: Duration::from_secs(0),
                    supervision_timeout: Duration::from_secs(5),
                })` 
```
to update connections parameters. But this function always returns `BleHost(Disconnected)`. I'm not sure that whether the connection parameters are updated, but according to 

https://github.com/embassy-rs/trouble/blob/d17132618413f0fa840fee32af4ab23ca125bec2/host/src/connection.rs#L415-L417

it's because that the conn handler is not recognized by the controller, which seems wrong. The data exchange of the connection runs well before and after calling `update_connection_params`

HaoboGu:
fixes #338

### DIFF
--- a/host/src/channel_manager.rs
+++ b/host/src/channel_manager.rs
@@ -109,7 +109,7 @@ impl<'d, P: PacketPool> ChannelManager<'d, P> {
         }
     }
 
-    fn next_request_id(&self) -> u8 {
+    pub(crate) fn next_request_id(&self) -> u8 {
         self.state.borrow_mut().next_request_id()
     }
 
--- a/host/src/connection.rs
+++ b/host/src/connection.rs
@@ -1,6 +1,6 @@
 //! BLE connection.
 
-use bt_hci::cmd::le::{LeConnUpdate, LeReadPhy, LeSetPhy};
+use bt_hci::cmd::le::{LeConnUpdate, LeReadLocalSupportedFeatures, LeReadPhy, LeReadRemoteFeatures, LeSetPhy};
 use bt_hci::cmd::status::ReadRssi;
 use bt_hci::controller::{ControllerCmdAsync, ControllerCmdSync};
 use bt_hci::param::{
@@ -18,6 +18,7 @@ use crate::pdu::Pdu;
 use crate::prelude::{AttributeServer, GattConnection};
 #[cfg(feature = "security")]
 use crate::security_manager::BondInformation;
+use crate::types::l2cap::ConnParamUpdateReq;
 use crate::{BleHostError, Error, Identity, PacketPool, Stack};
 
 /// Connection configuration.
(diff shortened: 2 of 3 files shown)


HaoboGu:
`LeReadRemoteFeatures` is not added because I have yet because I have no idea about how to get the result of `LeReadRemoteFeatures`, which is a `LeReadRemoteFeaturesComplete` event.

And, there is `Read Remote Supported Features` command, I'm not sure which one should be used

HaoboGu:
Another weird thing is, when acting as central, `LeReadLocalSupportedFeatures` cmd still returns `supports_conn_parameters_request_procedure = false`. But using `LeConnUpdate` to update the connection parameters runs well, so I add the `role == LeConnRole::Central` condition. Maybe I somehow misunderstood the spec, but it works for now.

host/src/channel_manager.rs
```diff
@@ -109,7 +109,7 @@ impl<'d, P: PacketPool> ChannelManager<'d, P> {
         }
     }
 
-    fn next_request_id(&self) -> u8 {
+    pub(crate) fn next_request_id(&self) -> u8 {
```
lulf: I would prefer if this was crate internal to avoid it leaking into other places

host/src/connection.rs
```diff
@@ -304,27 +305,54 @@ impl<'stack, P: PacketPool> Connection<'stack, P> {
...
-            .await
-        {
-            Ok(_) => Ok(()),
-            Err(BleHostError::BleHost(crate::Error::Hci(bt_hci::param::Error::UNKNOWN_CONN_IDENTIFIER))) => {
-                Err(crate::Error::Disconnected.into())
+        // First, check the local supported features to ensure that the connection update is supported.
+        let features = stack.host.command(LeReadLocalSupportedFeatures::new()).await?;
+        // TODO: How to read from `LeReadRemoteFeaturesComplete` event?
+        // let remote_features = stack.host.async_command(LeReadRemoteFeatures::new(handle)).await?;
+        if features.supports_conn_parameters_request_procedure() || self.role() == LeConnRole::Central {
```
lulf: What you need to do is issue the command, then parse the event in the rx future and somehow report that back to this function. I'm fine leaving this as is for now.

HaoboGu: Okay, I removed it and did some cleanup

host/src/channel_manager.rs
HaoboGu: Fixed

PR #412, opened 2025-06-19
Scan window interval not being passed to LeSetScanParams

ericschaal:
Hi,

Noticed a small typo in `Scanner`. Scanning `interval` duration was passed downstream for both `le_scan_interval` and `le_scan_window`.


lulf (approved):
Thanks!

Issue #336, opened 2025-04-01
Controller for regular old linux


Ciantic:
I'm planning to use bt-hci with regular old linux, like in Raspberry PI 4 or Zero.

It should be doable, one just opens HCI socket like this:

```rust
let fd = unsafe {
        libc::socket(
            libc::AF_BLUETOOTH,
            libc::SOCK_RAW | libc::SOCK_CLOEXEC,
            BTPROTO_HCI,
        )
    };
    if fd < 0 {
        return Err(Box::new(io::Error::last_os_error()));
    }

    // Bind to hci0 (device index 0) on the raw HCI channel.
    let mut addr: sockaddr_hci = unsafe { mem::zeroed() };
    addr.hci_family = libc::AF_BLUETOOTH as u16;
    addr.hci_dev = 0; // Use hci0; adjust if needed.
    addr.hci_channel = HCI_CHANNEL_RAW;
    let ret = unsafe {
        libc::bind(
            fd,
            &addr as *const _ as *const libc::sockaddr,
            mem::size_of::<sockaddr_hci>() as libc::socklen_t,
        )
    };
```

(Taken shamelessly from here: https://github.com/kevinisabelle/raspigamecontroller/commit/8e2cbd8c41b82d53cc1e226096792056c9c4b8a8 )

I'm not entirely sure but it should then just interacting with `write` and `read`, it should work with most Linux installations with bluez running or not.

I know there is one entire C codebase that works like this: https://github.com/petzval/btferret/blob/main/btferret.c

lulf:
It should work then I think if you implement [this trait](https://docs.rs/bt-hci/0.2.1/bt_hci/transport/trait.Transport.html).

Ciantic:
I might have opened this issue in wrong repo, if one were to create socket based transport, should it be merged to bt-hci or trouble?

I wanted to contribute or suggest a feature for regular old Linux socket transport. It would generally make bt-hci more useful because it could be used in Linux projects for Bluetooth. This socket based transport could be used wit embassy's std examples.

I now see that SerialTransport is inside the bt-hci. However, in my mind, bt-hci is just the protocol definition and shouldn't have implementation details like opening a socket.

lulf:
> I might have opened this issue in wrong repo, if one were to create socket based transport, should it be merged to bt-hci or trouble?
> 
> **Edit** I do think this is right repo, I just make an example (like others are) with socket transport.
> 

[...]

PR #420, opened 2025-07-13
Add example using Linux HCI sockets

logan12358:
This example addresses #336.

The implementation isn't massive so here it's embedded in the example, but it could be moved out if that's preferred.

I'm a beginner to Rust, so open to any feedback and happy to make any changes.

Some comments:
* Quick start at `examples/linux/README.md`.
* `BigAlloc` from the `serial-hci` example is moved into `apps` to be shared with the new example.
* I seem to only have one device supporting DLE, so I haven't tested the high throughput examples end-to-end. The other examples all seem to work.
* With the `security` feature enabled, _only_ the `*_sec` examples run (the others panic due to unseeded RNG). I think this affects all of the examples, so I'm not sure if it's much of an issue.
* `std::io::{Read, Write}` just impl'd to use for implementing `tokio::io::{AsyncRead, AsyncWrite}`. The `std::io` implementations could be inlined to shorten things, but it seems kind of nice to keep all of the sync and potentially unsafe code together?
* We could _almost_ use `bt_hci::transport::SerialTransport`, but the Linux socket expects a whole HCI packet in a single `write` (rather than a bunch of separate writes). If `SerialTransport` was changed to flush after writing, we could possibly simplify things with a buffered writer and `embedded-io-adapters`.
* It is possible to open a "raw" rather than "user" channel (and set appropriate filters), to operate alongside other Linux users like `bluetoothd`. I saw some weird behaviours when doing this and I don't think it makes sense.

Thanks to @Ciantic for the starting point in the linked issue.


logan12358:
> `std::io::{Read, Write}` just impl'd to use for implementing `tokio::io::{AsyncRead, AsyncWrite}`. The `std::io` implementations could be inlined to shorten things, but it seems kind of nice to keep all of the sync and potentially unsafe code together?

I talked myself around on this, it's simpler without the indirection through `std::io::{Read, Write}` and they're not really useful for anything else. Updating the PR.

lulf:
Great, thanks for the PR!

> This example addresses #336.
> 
> The implementation isn't massive so here it's embedded in the example, but it could be moved out if that's preferred.
> 
> I'm a beginner to Rust, so open to any feedback and happy to make any changes.
> 
> Some comments:
> 
> * Quick start at `examples/linux/README.md`.
> * `BigAlloc` from the `serial-hci` example is moved into `apps` to be shared with the new example.
> * I seem to only have one device supporting DLE, so I haven't tested the high throughput examples end-to-end. The other examples all seem to work.
> * With the `security` feature enabled, _only_ the `*_sec` examples run (the others panic due to unseeded RNG). I think this affects all of the examples, so I'm not sure if it's much of an issue.

All of the above sounds good.

> * `std::io::{Read, Write}` just impl'd to use for implementing `tokio::io::{AsyncRead, AsyncWrite}`. The `std::io` implementations could be inlined to shorten things, but it seems kind of nice to keep all of the sync and potentially unsafe code together?
> * We could _almost_ use `bt_hci::transport::SerialTransport`, but the Linux socket expects a whole HCI packet in a single `write` (rather than a bunch of separate writes). If `SerialTransport` was changed to flush after writing, we could possibly simplify things with a buffered writer and `embedded-io-adapters`.

Hmm, that's intereresting, because I don't think that follows the read/write POSIX guarantees. Do you have a reference for why/what hci socket does?

I originally thought the read half would keep a buffer that it read into until it got a full packet.

Another issue I think is that the transport async implementations are blocking, and they should not be. I think the proper way to handle that is to use libc select on the socket to check if it is ready for read/write and do the corresponding Poll::Pending in those implementations.

logan12358:
> > * We could _almost_ use `bt_hci::transport::SerialTransport`, but the Linux socket expects a whole HCI packet in a single `write` (rather than a bunch of separate writes). If `SerialTransport` was changed to flush after writing, we could possibly simplify things with a buffered writer and `embedded-io-adapters`.

> Hmm, that's intereresting, because I don't think that follows the read/write POSIX guarantees. Do you have a reference for why/what hci socket does?
> 
> I originally thought the read half would keep a buffer that it read into until it got a full packet.

Since the socket type is `SOCK_RAW` (not a stream), I think it is expected that each read/write syscall deals with a complete HCI packet. [^1] I haven't found a reference explicitly stating this, and don't know much about Linux bluetooth internals, but [this length check](https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/net/bluetooth/hci_sock.c#L1816) in the `sendmsg` implementation suggests it expects to deal with a complete packet.

I might have explained the "problem" (not a very important one IMO) here poorly. My point is just that we have to [buffer packets for writing](https://github.com/embassy-rs/trouble/blob/ee2313f9b9c91189a65d1516501b61299e998c53/examples/linux/src/lib.rs#L155), rather than using `HostToControllerPacket::write_hci_async` directly into the socket, since that would make multiple small write calls and doesn't (can't?) give an indication of when to flush.

If `bt_hci::transport::SerialTransport` called `embedded_io_async::Write::flush` after its [write](https://docs.rs/bt-hci/latest/src/bt_hci/transport.rs.html#103), so we could glue together `SerialTransport` + `tokio::io::BufWriter` + `embedded_io_adapters::tokio_1::FromTokio` rather than implementing our own `Transport` in this example. The more I think about it the less I'm convinced this would be any "nicer", though...

> Another issue I think is that the transport async implementations are blocking, and they should not be. I think the proper way to handle that is to use libc select on the socket to check if it is ready for read/write and do the corresponding Poll::Pending in those implementations.

I'm pretty sure (but might be missing some subtleties) that this async implementation is nonblocking. The socket is opened as `SOCK_NONBLOCK` so operations will fail with `EAGAIN` if they would block. Then, [`AsyncFd`](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html) does the heavy lifting of polling for readiness and turning "would block" errors into `Pending`.

[^1]: From `socket(2)`:
    > SOCK_DGRAM and SOCK_RAW sockets allow sending of datagrams to correspondents named in sendto(2) calls.  Datagrams are generally received with recvfrom(2), which returns the next datagram along with the address of its sender.

lulf:
> > > * We could _almost_ use `bt_hci::transport::SerialTransport`, but the Linux socket expects a whole HCI packet in a single `write` (rather than a bunch of separate writes). If `SerialTransport` was changed to flush after writing, we could possibly simplify things with a buffered writer and `embedded-io-adapters`.
> 
> > Hmm, that's intereresting, because I don't think that follows the read/write POSIX guarantees. Do you have a reference for why/what hci socket does?
> > I originally thought the read half would keep a buffer that it read into until it got a full packet.
> 
> Since the socket type is `SOCK_RAW` (not a stream), I think it is expected that each read/write syscall deals with a complete HCI packet. [1](#user-content-fn-1-95049047b396fc17587c944eaeec0dc5) I haven't found a reference explicitly stating this, and don't know much about Linux bluetooth internals, but [this length check](https://github.com/torvalds/linux/blob/155a3c003e555a7300d156a5252c004c392ec6b0/net/bluetooth/hci_sock.c#L1816) in the `sendmsg` implementation suggests it expects to deal with a complete packet.
> 
> I might have explained the "problem" (not a very important one IMO) here poorly. My point is just that we have to [buffer packets for writing](https://github.com/embassy-rs/trouble/blob/ee2313f9b9c91189a65d1516501b61299e998c53/examples/linux/src/lib.rs#L155), rather than using `HostToControllerPacket::write_hci_async` directly into the socket, since that would make multiple small write calls and doesn't (can't?) give an indication of when to flush.
> 
> If `bt_hci::transport::SerialTransport` called `embedded_io_async::Write::flush` after its [write](https://docs.rs/bt-hci/latest/src/bt_hci/transport.rs.html#103), so we could glue together `SerialTransport` + `tokio::io::BufWriter` + `embedded_io_adapters::tokio_1::FromTokio` rather than implementing our own `Transport` in this example. The more I think about it the less I'm convinced this would be any "nicer", though...
> 
> > Another issue I think is that the transport async implementations are blocking, and they should not be. I think the proper way to handle that is to use libc select on the socket to check if it is ready for read/write and do the corresponding Poll::Pending in those implementations.
> 
> I'm pretty sure (but might be missing some subtleties) that this async implementation is nonblocking. The socket is opened as `SOCK_NONBLOCK` so operations will fail with `EAGAIN` if they would block. Then, [`AsyncFd`](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html) does the heavy lifting of polling for readiness and turning "would block" errors into `Pending`.

Maybe I got the wrong impression from this code (not familiar with AsyncFd and the ready! macro). 

```
        loop {
            let mut guard = ready!(fd.poll_read_ready_mut(cx))?;

```

Does the above part mean that it will return Poll::Pending if there is nothing to read?

```
            let unfilled = buf.initialize_unfilled();
            match guard.try_io(|inner| {
                let ret = unsafe { libc::read(inner.as_raw_fd(), unfilled.as_mut_ptr().cast(), unfilled.len()) };
                usize::try_from(ret).map_err(|_try_from_int_error| io::Error::last_os_error())
            }) {
                Ok(result) => {
                    return Poll::Ready(result.map(|len| buf.advance(len)));
                }
                Err(_would_block) => {}
            }
```

It is the Err(_would_block) that confuses me a little. Shouldn't it return Poll::Pending in that case?


> ## Footnotes
> 
>     1. From `socket(2)`:
>        > SOCK_DGRAM and SOCK_RAW sockets allow sending of datagrams to correspondents named in sendto(2) calls.  Datagrams are generally received with recvfrom(2), which returns the next datagram along with the address of its sender.
>        
>        
>        [↩](#user-content-fnref-1-95049047b396fc17587c944eaeec0dc5)

logan12358:
> Maybe I got the wrong impression from this code (not familiar with AsyncFd and the ready! macro).
> 
> ```
>         loop {
>             let mut guard = ready!(fd.poll_read_ready_mut(cx))?;
> ```
> 
> Does the above part mean that it will return Poll::Pending if there is nothing to read?

Yup, [ready](https://doc.rust-lang.org/stable/std/task/macro.ready.html) just returns a Pending or evaluates to the ready value.

> 
> ```
>             let unfilled = buf.initialize_unfilled();
>             match guard.try_io(|inner| {
>                 let ret = unsafe { libc::read(inner.as_raw_fd(), unfilled.as_mut_ptr().cast(), unfilled.len()) };
>                 usize::try_from(ret).map_err(|_try_from_int_error| io::Error::last_os_error())
>             }) {
>                 Ok(result) => {
>                     return Poll::Ready(result.map(|len| buf.advance(len)));
>                 }
>                 Err(_would_block) => {}
>             }
> ```
> 
> It is the Err(_would_block) that confuses me a little. Shouldn't it return Poll::Pending in that case?

By continuing, I think `Err(_would_block)` just loops around and returns Pending at `ready!` (with the waker/context setup from `poll_read_ready_mut`). These implementations pretty much come from the AsyncFd [example](https://docs.rs/tokio/latest/tokio/io/unix/struct.AsyncFd.html#examples).

lulf (approved):
Good points, I think this is good to merge then. What do you think about splitting the linux controller implementation out in a separate crate? It could have uses outside trouble as well. I would propose moving it into a repo embassy-rs/bt-hci-linux or something.

logan12358:
> Good points, I think this is good to merge then. What do you think about splitting the linux controller implementation out in a separate crate? It could have uses outside trouble as well. I would propose moving it into a repo embassy-rs/bt-hci-linux or something.

Absolutely, that makes sense. I don't know if there would be any extra concerns to implying it's more than an example. Naming (for `Socket` and `Transport`) and the error type might want a bit more thought?

Thanks for your time looking through this!

lulf:
> > Good points, I think this is good to merge then. What do you think about splitting the linux controller implementation out in a separate crate? It could have uses outside trouble as well. I would propose moving it into a repo embassy-rs/bt-hci-linux or something.
> 
> Absolutely, that makes sense. I don't know if there would be any extra concerns to implying it's more than an example. Naming (for `Socket` and `Transport`) and the error type might want a bit more thought?
> 

Yeah, that sounds good. I'll merge this for now anyway (the example-tests failing is not related to this PR)

> Thanks for your time looking through this!

iwanders:
>  I don't know if there would be any extra concerns to implying it's more than an example.

For what it's worth, Linux HCI transport has been working _really_ well for me. It makes things much easier to be able to test on a platform that is not a microcontroller. Thank you for making this, it's been super helpful!

> What do you think about splitting the linux controller implementation out in a separate crate? It could have uses outside trouble as well.

It is actually possible to directly consume this code without a local checkout, copy pasting or creating a crate. In `cargo.toml` use the git repository directly ([docs for that functionality](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-git-repositories)), I pinned it at a specific commit here:
```toml
trouble-linux-examples = { git = "https://github.com/embassy-rs/trouble", rev = "1e6f2d743857ab6888a8a59199ad0c186bbf38c7" }
```
And then, just like in the [example](https://github.com/embassy-rs/trouble/blob/366ee88a2aa19db11eb0707c71d797156abe23f5/examples/linux/src/bin/ble_bas_peripheral.rs#L3) you can just import the tranport:
```rust
use trouble_linux_examples::Transport;
````

I'm not sure how well-known this trick is though, I'm happy to file a PR to add it to the [linux example readme](https://github.com/embassy-rs/trouble/tree/366ee88a2aa19db11eb0707c71d797156abe23f5/examples/linux#linux-hci-socket-example) if that's considered reasonable?

lulf:
I don't think we should advocate that, because then the trouble_linux_examples becomes a public API that people depend on, and I think we'd want to be able to change/move things around without worrying about that.

I can take the steps to move this into an external crate though, so created https://github.com/embassy-rs/trouble/issues/444 to track that.

PR #424, opened 2025-07-15
Implement just works, numeric comparison and pass key entry

AndWass:
Apologies for this drive-by pull request. I went down a bit of a rabbit hole on this and figured I had landed in a state where it might be interesting for upstream to take a look. There are quite alot of changes but I think this leaves the security manager in a much better state and is actually usable for a broader range of applications.

@Erik1000 I saw you commented on the current state of the security manager in an issue, maybe this is interesting to you too?

## Current state

This is my analysis of the current state, but it might be incomplete.

As far as I can see the current security manager lacks support for anything but the most basic just-works pairing. There is no way for the user to specify the IO capabilities that is then used to select the pairing method, meaning that Just works will always be used. There is also no way to confirm or inject pass keys in the current implementation.

It is also only possible to check if the connection is encrypted, there is no way to check if it is authenticated (which it never is currently).

The security manager in its current state doesn't really lend itself to be evolved towards multiple simultaneous pairings.

## Implemented

  - [x] Just works
  - [x] Numeric comparison
  - [x] Pass key entry
    - Keyboard only and 1 display 1 keyboard
  - [x] Differentiate between authenticated and unauthenticated pairings
  - [ ] Out of band
  - [x] Bonding
    - The basic bonding capability is preserved but bonding must now be explicitly enabled on each `Connection` through `conn.set_bondable(true)`. This must be done before pairing. 
  - [x] Design towards future multiple simultaneous pairings.
  - [ ] Key exchange in phase 3
  - [ ] IRK support
  - [x] Re-encrypt using bonded LTK
  - [x] Pairing timeout

## Design decisions

At the moment the LTK is stored within the security manager, but in my view it could probably also be stored in each `ConnectionStorage` which could probably simplify some parts, at the expense of slightly more memory being used.

~~The bonded connection event is sent if both peripheral and central indicates that it wants to bond. This must be enabled before pairing by a call to `conn.set_bondable(true)`.~~

Bond information is now an optional associated data to the `PairingComplete` connection event. If `conn.set_bondable(true)` hasn't been called, this will always `None`, otherwise it depends on if both peripheral and central are bondable or not.

## Future

Implement out of band: Should be straight forward, should only touch the state machines and adding a connection event to ask for OOB information.

Phase 3 key exchange: Not implemented.

IRK support: Not implemented.


host/src/security_manager/types.rs
```diff
@@ -280,73 +281,69 @@ impl defmt::Format for Command {
...
-        write!(f, "{}", self.as_ref())
     }
 }
 
 #[cfg(feature = "defmt")]
-impl defmt::Format for IoCapabilities {
+impl defmt::Format for PassKey {
     fn format(&self, fmt: defmt::Formatter) {
-        defmt::write!(fmt, "{}", self.as_ref())
+        if self.0 < 10 {
```
HaoboGu: Just use: https://defmt.ferrous-systems.com/hints#zero-padding

host/src/lib.rs
```diff
@@ -665,25 +665,4 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
     pub fn log_status(&self, verbose: bool) {
         self.host.log_status(verbose);
     }
-
-    #[cfg(feature = "security")]
```
HaoboGu: Those functions should not be removed, the BLE host needs to provide a way for retrieving and loading bond information which is saved on the user-side

host/src/security_manager/types.rs
AndWass: Thanks! Will update

host/src/lib.rs
AndWass: I can certainly add that back in. In the future I want to explore alternative ways of handling bonding storage with the goal of removing the `BOND_COUNT` generic argument, but until then i will restore the original capability

AndWass: @HaoboGu I have restored the functionality but bonding must be enabled on a connection by setting bondable to true: `conn.set_bondable(true)` before pairing is initiated. Bonding will only be performed if both central and peripheral sets this, otherwise the `Bonded` event will not be generated.

HaoboGu: Can the `bondable` be set automatically according to the advertising type? Or could you add an example for creating bondable connection and saving & loading the bond information?

examples/apps/src/ble_advertise.rs
```diff
@@ -19,7 +20,7 @@ where
     info!("Our address = {:?}", address);
 
     let mut resources: HostResources<DefaultPacketPool, 0, 0> = HostResources::new();
-    let stack = trouble_host::new(controller, &mut resources).set_random_address(address);
+    let stack = trouble_host::new(controller, &mut resources, IoCapabilities::NoInputNoOutput).set_random_address(address);
```
HaoboGu: This is a breaking change of current API. 

Maybe

```rust
trouble_host::new(controller, &mut resources).set_random_address(address).set_io_capabilities(IoCapabilities::NoInputNoOutput);
```

is better?

HaoboGu:
Since @blueluna is the original author of the security manager part, I think @blueluna might have more comments.

host/src/lib.rs
AndWass: Bondable isn't detectable by advetisement type. It is a capability where a device should only be bondable if it is possible to store bonding info in NVM. And it should only be stored/reused if both central and peripheral supports bonding.

In the current implementation the central handling is wrong since it will try to enable encryption without pairing if it has successfully paired previously, even if the peripheral device doesnt support bonding (and thus wont store the LTK). 

Likewise it is wrong as a peripheral to support bonding if its not possible to store bonding info since it would prevent reemcrypting the communication if the central has stored the LTK but peripheral hasnt. 

I think it would make sense to have a (stack-)global default that can also be overridden per connection. This is also what Zephyr does.

I can also add an example.

examples/apps/src/ble_advertise.rs
AndWass: Yah I think defaulting to no input output with a separate io capability setter makes sense and would be less breaking.

host/src/lib.rs
HaoboGu: thanks for the explanation!

blueluna:
> Since @blueluna is the original author of the security manager part, I think @blueluna might have more comments.

I am on leave, but had a quick look. looks good to me.

examples/apps/src/ble_bas_peripheral_auth.rs
```diff
@@ -0,0 +1,237 @@
+use embassy_futures::join::join;
+use embassy_futures::select::{select, Either};
+use embassy_time::Timer;
+use embedded_hal_async::digital::Wait;
+use rand_core::{CryptoRng, RngCore};
+use trouble_host::IoCapabilities;
+use trouble_host::prelude::*;
```
lulf: Please include the IoCapabilities in the prelude.

host/src/lib.rs
```diff
@@ -12,15 +12,16 @@ use bt_hci::cmd::status::ReadRssi;
...
-#[cfg(feature = "security")]
-use heapless::Vec;
 use rand_core::{CryptoRng, RngCore};
 
 use crate::att::AttErrorCode;
 use crate::channel_manager::ChannelStorage;
 use crate::connection_manager::ConnectionStorage;
 #[cfg(feature = "security")]
-pub use crate::security_manager::{BondInformation, IdentityResolvingKey, LongTermKey};
+pub use crate::security_manager::{IdentityResolvingKey, LongTermKey, BondInformation};
```
lulf: I think think these should maybe be available in the trouble_host::prelude:: and trouble_host::security:: API-wise.

lulf (approved):
LGTM. I agree with @HaoboGu on adding a set_io_capabilities() with default NoInputNoOutput instead of adding it to the new() func.

host/src/lib.rs
AndWass: Added these to the prelude now together with `IoCapabilities`

AndWass:
Thanks for the reviews! I have added new commits with prelude updates and removed `IoCapabilities` from `trouble_host::new` and added a new `set_io_capabilities` function to set the IO capabilities.

One outstanding question is if there should be a host-global flag for bondable instead of having to manually set this on each connection. This would allow something like

```
trouble_host::new(controller, &mut resources)
        .set_random_address(address)
        .set_random_generator_seed(random_generator)
        .set_default_bondable(true);
```

which would then make all connections default to bondable, with ability to disable on a per-connection basis. In my opinion I think this would be a good idea.

HaoboGu:
> One outstanding question is if there should be a host-global flag for bondable instead of having to manually set this on each connection. This would allow something like
> 
> ```
> trouble_host::new(controller, &mut resources)
>         .set_random_address(address)
>         .set_random_generator_seed(random_generator)
>         .set_default_bondable(true);
> ```
> 
> which would then make all connections default to bondable, with ability to disable on a per-connection basis. In my opinion I think this would be a good idea.

The bondable defaults to `false`, right? If so, I think it worth to have some documentation. 

And also the behavior of getting/setting bond information should be changed according to the `bondable` flag, because when `bondable` is false, getting/setting bond information becomes meaningless. Information or guide should be provided that users could follow to enable bondable and set NVM for storing bond information

AndWass:
> The bondable defaults to `false`, right? If so, I think it worth to have some documentation.
> 

Yes, I can add some documentation regarding this.

> And also the behavior of getting/setting bond information should be changed according to the `bondable` flag, because when `bondable` is false, getting/setting bond information becomes meaningless. Information or guide should be provided that users could follow to enable bondable and set NVM for storing bond information

Currently, the bondable flag is only on a per-connection basis and defaults to `false`.

The current behaviour when bondable is false is the following: 

## Central

  * Pairing initated **or** security request received
    * Bond stored:
      * Enable encryption with stored LTK, report success or failure to user
      * On disconnect, do **not** remove the stored bond (it might be some kind of pre-provisioned bond for instance).
    * Bond not stored:
      * Initiate pairing, not bondable, using IO capabilities to select pairing method
      * Report success or failure to user
      * On disconnect remove the LTK used by the connection

## Peripheral

  * Initated by user: Send security request to central not bondable.
  * Pairing request received by central:
    * Send pairing response not bondable. Use IO capabilities to select pairing method
    * Report success or failure to user
    * On disconnect remove the LTK used by the connection.
    * __In case a bond already exists, should we reject the pairing request?__ I think Zephyr allows the pairing to continue, but more advanced use-cases can be supported since Zephyr allows for confirming the pairing by the user, so the application could perform its own rejection on-top of the default behaviour.

When bondable is true the same flow is followed, but security/pairing request and pairing response set the device as bondable. ~~After successful pairing is reported to the user, a `Bonded` event is also reported but otherwise it is the same flow.~~ After a successful pairing, the success and bond information is reported in the `bond` associated data of the `PairingComplete` event.

The `bondable` flag should have no bearing on whether you allow injecting bonds or not. Consider the following use-cases:

  * You pre-provision the LTK and don't accept bonding otherwise. In that case you would use `stack.add_bond_information` to add the pre-provisioned LTK to allow establishing encrypted communication.
  * You allow a very limited number of bonds, and when full you no longer allow new bonds. Again the user must be able to use `add_bond_information` even if the device isn't bondable any longer.

My thinking is that we have a default-bondable flag as well, that the per-connection `bondable` defaults to. That way the user won't have to set `bondable` to true on each connection, instead it could just set the default bondable flag.

I'll try and add some complete bondable example as well.

examples/apps/src/ble_bas_peripheral_auth.rs
AndWass: Done

examples/apps/src/ble_advertise.rs
AndWass: Done

AndWass:
Latest changes:
Bond information is now an optional associated data to the PairingComplete connection event. If conn.set_bondable(true) hasn't been called, this will always None, otherwise it depends on if both peripheral and central are bondable or not.

I have added some documentation on the `set_bondable` method to clarify when and how bond information is sent.

I have also added examples of both peripheral and central that can store bond information for a single peer.

I have not added the default-bondable flag I mentioned in a comment above. At the moment I don't have enough knowledge of the connection procedure to know where to inject this. If needed I can dive into this, or this could maybe come in a future PR?

host/src/lib.rs
```diff
@@ -618,6 +622,19 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
         }
         self
     }
+    /// Set the IO capabilities used by the security manager.
+    ///
+    /// Only relevant if the feature `security` is enabled.
+    pub fn set_io_capabilities(self, io_capabilities: IoCapabilities) -> Self {
+        #[cfg(feature = "security")]
```
lulf, line 629: Would it make sense to guard the entire API behind the security feature?

AndWass: Yah probably. I just followed the same style that was used by `set_random_generator_seed` so no real preference in my opinion.

PR #426, opened 2025-07-21
Update esp-hal

bugadani:
I hope you don't mind the rust-toolchain bump 🙃

### DIFF
--- a/examples/esp32/src/bin/ble_bas_peripheral_sec.rs
+++ b/examples/esp32/src/bin/ble_bas_peripheral_sec.rs
@@ -2,11 +2,14 @@
 #![no_main]
 
 use embassy_executor::Spawner;
-use esp_hal::{clock::CpuClock, timer::timg::TimerGroup};
+use esp_hal::clock::CpuClock;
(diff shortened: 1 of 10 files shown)


examples/esp32/src/bin/ble_bas_peripheral_sec.rs
```diff
@@ -17,7 +20,7 @@ async fn main(_s: Spawner) {
 
     let mut rng = esp_hal::rng::Trng::new(peripherals.RNG, peripherals.ADC1);
```
HaoboGu, line 21: I'm wondering what's the difference between 

```rust
esp_hal::rng::Rng::new(peripherals.RNG)
```

and 

```rust
esp_hal::rng::Trng::new(peripherals.RNG, peripherals.ADC1)
```
?

I found in different files, the `rng` initialization is different

bugadani: Trng enables an additional entropy source using the ADC.

HaoboGu: Thanks!

PR #427, opened 2025-07-22
L2CAP Peripheral iOS/macOS Example

horho77:
This PR adds a simple iOS/macOS L2CAP peripheral example that echoes received data back to the central device.

- The example uses a GATT service to [dynamically publish the PSM](https://developer.apple.com/documentation/corebluetooth/cbuuidl2cappsmcharacteristicstring) on channels compatible with the iOS/macOS CoreBluetooth library

- [L2CapDemo](https://github.com/paulw11/L2CapDemo) project can be used for basic testing

I'm not sure this is unique enough to be justified to be in examples. Let me know.
Thanks 😊

### DIFF
--- a/examples/apps/src/ble_l2cap_peripheral_ios.rs
+++ b/examples/apps/src/ble_l2cap_peripheral_ios.rs
@@ -10,6 +10,7 @@ const CONNECTIONS_MAX: usize = 1;
 const L2CAP_CHANNELS_MAX: usize = 3; // Signal + att + CoC
 
 /// L2CAP PSM to be published to the client
+/// Only values in range 0x0040 - 0x00FF are accepted by iOS/macOS clients.
 const L2CAP_PSM: u16 = 0x00C0;
 
 /// GATT Server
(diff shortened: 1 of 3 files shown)


lulf:
Could you perhaps modify the l2cap peripheral example to publish the gatt service instead?

horho77:
> Could you perhaps modify the l2cap peripheral example to publish the gatt service instead?

Are you suggesting to modify both peripheral & central L2CAP examples to use the "dynamic PSM" by default? Just to make sure I understand your comment right....

lulf:
> > Could you perhaps modify the l2cap peripheral example to publish the gatt service instead?
> 
> Are you suggesting to modify both peripheral & central L2CAP examples to use the "dynamic PSM" by default? Just to make sure I understand your comment right....

Yes, that's correct. If structured in a way that makes it possible to "rip out" the gatt part if needed. The trouble-host features needed to use gatt are already there so I don't think it will make them too noise. And it's easier to maintain than another example.

PR #431, opened 2025-07-30
Use L2CAP signaling for connection update

timokroeger:
Follow up to #411.

From BT Core Spec 6.1 [Vol 3] Part B, Section 5.1.1. Connection Update procedure "If the Central, the Peripheral, or both do not support the Connection Parameters Request procedure, then the Central shall send an LL_CONNECTION_UPDATE_IND PDU (the Peripheral shall not send this PDU) while the Peripheral shall use the L2CAP LE Signaling channel."


lulf (approved):
Thanks!

PR #437, opened 2025-08-12
made scan_command_state.cancel in scan_ext use the right argument

Tomer-Eliahu:
I believe this was incorrectly set to false. Now if this fires, it cancels extended scanning (based on reading host.rs) and not regular scanning (one has no impact on the other from what I understand).


lulf (approved):
Good catch, thank you!

PR #442, opened 2025-09-01
Don't depend on bt-hci/embassy-time

bugadani:
This is part of my master plan to remove embassy-time from bt-hci. bt-hci depends on embassy-time for the `From` conversion, which is exactly 2 (or 6, depending on how you count it) lines of code and just means bt-hci needs to be updated more often than strictly necessary.


lulf (approved):
Agreed with your plan, thanks!

PR #443, opened 2025-09-02
Service discovery may fail if previous service ends on 16 aligned boundary.

iwanders:
Hi, I ran into a situation where the order of my services in the `#[gatt_server]`-derived struct made one of my services disappear.

I don't think linking to my code will make things clearer, since that's very non-minimal (and messy), but the rough situation that I had:
```rust
#[gatt_server]
struct Server {
    accessory_information: AccessoryInformationService,
    protocol: ProtocolInformationService,
    pairing: PairingService,
    lightbulb: LightbulbService,
}
```
Which caused the `LightbulbService` to not be present according to nRF Connect (only the first three apeared). I first started ruling out my code, because I manually handle the gatt read and write events, but use the derive macro's to create the gatt server as a facade.

After looking manually at the requests coming in, and what was handled, everything looked fine, but for some reason the last `ReadByGroupType` just didn't return something, I thought maybe I ran out of space in my gatt table or something, so I tried moving the lightbulb service up such that it wasn't the last one anymore:
```rust
#[gatt_server]
struct Server {
    accessory_information: AccessoryInformationService,
    lightbulb: LightbulbService,
    protocol: ProtocolInformationService,
    pairing: PairingService,
}
```
Which worked (all four services appeared), so the next question was why does this work, and the other situation doesn't.

Eventually I got the following prints, mostly from the [handle_read_by_group_type_req](https://github.com/embassy-rs/trouble/blob/366ee88a2aa19db11eb0707c71d797156abe23f5/host/src/attribute_server.rs#L422) function:

```
WARN  trouble_host::attribute_server] Looking for group: Uuid16([0, 28]) between 75 and 65535
DEBUG trouble_host::attribute_server] [read_by_group] found! Uuid16([0, 28]) 80
DEBUG trouble_host::attribute_server] last_handle_in_group: 96
DEBUG trouble_host::attribute_server] read_attribute_data: Ok(16)
TRACE trouble_host::host] [host] granted send packets = 1, len = 30
TRACE trouble_host::host] [host] sent acl packet len = 26
TRACE trouble_host::host] [host] inbound l2cap header channel = 4, fragment len = 7, total = 7
INFO  main_ble::ble_bas_peripheral] [gatt-attclient]: ReadByGroupType { start: 97, end: 65535, group_type: Uuid16([0, 40]) }
INFO  main_ble::ble_bas_peripheral] [gatt] other event
WARN  trouble_host::attribute_server] Looking for group: Uuid16([0, 28]) between 97 and 65535
WARN  trouble_host::attribute_server] [read_by_group] Dit not find attribute Uuid16([0, 28]) between 97  65535
```
Note that the `last_handle_in_group: 96`, and the nRF connect app (or iOS?) thus requests the 97th handle next to look for the next service.

When comparing this to a dump of the the entire table:
```
table handle 72 uuid Uuid128([9a, 93, 96, d7, bd, 6a, d9, b5, 16, 46, d2, 81, fe, f0, 46, dc]), last: 74
table handle 80 uuid Uuid16([0, 28]), last: 96
table handle 81 uuid Uuid16([3, 28]), last: 96
table handle 82 uuid Uuid128([d1, a0, 83, 50, 0, aa, d3, 87, 17, 48, 59, a7, 5d, e9, 4, e6]), last: 96
table handle 83 uuid Uuid16([3, 28]), last: 96
table handle 84 uuid Uuid128([91, 52, 76, bb, 26, 0, 0, 80, 0, 10, 0, 0, 4c, 0, 0, 0]), last: 96
table handle 85 uuid Uuid128([9a, 93, 96, d7, bd, 6a, d9, b5, 16, 46, d2, 81, fe, f0, 46, dc]), last: 96
table handle 86 uuid Uuid16([3, 28]), last: 96
table handle 87 uuid Uuid128([91, 52, 76, bb, 26, 0, 0, 80, 0, 10, 0, 0, 4e, 0, 0, 0]), last: 96
table handle 88 uuid Uuid128([9a, 93, 96, d7, bd, 6a, d9, b5, 16, 46, d2, 81, fe, f0, 46, dc]), last: 96
table handle 89 uuid Uuid16([3, 28]), last: 96
table handle 90 uuid Uuid128([91, 52, 76, bb, 26, 0, 0, 80, 0, 10, 0, 0, 4f, 0, 0, 0]), last: 96
table handle 91 uuid Uuid128([9a, 93, 96, d7, bd, 6a, d9, b5, 16, 46, d2, 81, fe, f0, 46, dc]), last: 96
table handle 92 uuid Uuid16([3, 28]), last: 96
table handle 93 uuid Uuid128([91, 52, 76, bb, 26, 0, 0, 80, 0, 10, 0, 0, 50, 0, 0, 0]), last: 96
table handle 94 uuid Uuid128([9a, 93, 96, d7, bd, 6a, d9, b5, 16, 46, d2, 81, fe, f0, 46, dc]), last: 96
table handle 96 uuid Uuid16([0, 28]), last: 109
table handle 97 uuid Uuid16([3, 28]), last: 109
table handle 98 uuid Uuid128([d1, a0, 83, 50, 0, aa, d3, 87, 17, 48, 59, a7, 5d, e9, 4, e6]), last: 109
table handle 99 uuid Uuid16([3, 28]), last: 109
```

The lightbulb service starts at line `table handle 96 uuid Uuid16([0, 28]), last: 109`.

I'm _not_ sure why the handle jumps from 94 to 96, but the issue is clear, the next service is on handle 96, and starting at 97 means we're searching beyond its start, which is why we get the unfortunate warning about this (I added this print):
```
WARN  trouble_host::attribute_server] [read_by_group] Dit not find attribute Uuid16([0, 28]) between 97  65535
```

With this information in hand, I set out to build a unit test that confirms this issue and fuzzes the length of a service to ensure we don't lose it if the length of its precursor falls on an unfortunate value.

This issue, I think (definitely out of my depth here), is in [the plus one](https://github.com/embassy-rs/trouble/blob/366ee88a2aa19db11eb0707c71d797156abe23f5/host/src/attribute.rs#L616) of the Attribute, without that `+ 1` things work as expected. I tried figuring out what the specification says and I think [this line](https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host/generic-attribute-profile--gatt-.html#:~:text=The%20End%20Group%20Handle%20is%20the%20handle%20of%20the%20last%20attribute%20within%20the%20service%20definition) is the best description, which states that it is the handle of the last attribute. I think the current implementation is the handle beyond that? 

I did not really find a unit test that was sending mock requests and verifying the expected responses. I think the integration tests in the `tests` directory are the only ones really doing this? But they require hardware from the looks of it?  I cobbled one together with some parts from the existing connection manager tests to get a connection to pass to the gatt server. I also couldn't find a parser for the response, so it's currently just pulling the two bytes we need for the next request out of a slice. It's not great, but the best I could come up with being unfamiliar with most of the codebase.

I hope this description is clear, happy to elaborate or incorporate feedback. (And thanks for creating "trouble"!)

### DIFF
--- a/host/src/attribute.rs
+++ b/host/src/attribute.rs
@@ -613,7 +613,6 @@ impl<'d, M: RawMutex, const MAX: usize> ServiceBuilder<'_, 'd, M, MAX> {
 
 impl<M: RawMutex, const MAX: usize> Drop for ServiceBuilder<'_, '_, M, MAX> {
     fn drop(&mut self) {
-        info!("dropping, current handle is {}", self.table.handle);
         let last_handle = self.table.handle;
         self.table.with_inner(|inner| {
             for item in inner.attributes[self.start..].iter_mut() {
--- a/host/src/attribute_server.rs
+++ b/host/src/attribute_server.rs
@@ -322,16 +322,6 @@ impl<'values, M: RawMutex, P: PacketPool, const ATT_MAX: usize, const CCCD_MAX: 
     pub fn new(
         att_table: AttributeTable<'values, M, ATT_MAX>,
     ) -> AttributeServer<'values, M, P, ATT_MAX, CCCD_MAX, CONN_MAX> {
-        att_table.iterate(|mut it| {
-            while let Some(att) = it.next() {
-                trace!(
-                    "Entire table handle {} uuid {:x?}, last: {}",
-                    att.handle,
-                    att.uuid,
-                    att.last_handle_in_group
-                );
-            }
-        });
         let cccd_tables = CccdTables::new(&att_table);
         AttributeServer {
             att_table,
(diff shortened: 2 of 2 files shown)


lulf (approved):
Thanks for the well written issue. I think you're absolutely right, the last_handle should not be the handle + 1, because that is beyond the last handle. 
 
And great to have a unit test like this as well!

PR #452, opened 2025-09-19
Implement a helper proc macro for defining UUIDs cleaner

davidlenfesty:
Just in time to miss the 0.3.0 release ;)

I wanted a helper so I could declare UUIDs with the string syntax nicely, in one place, so I made it.

Let me know if you want changes or there was a different way to do this already :P


lulf (approved):
Thanks!

PR #454, opened 2025-09-22
Fix: catch up esp32 'sec' examples to use esp-hal 1.0.0-rc0 TRng API

finalyards:
Noticed that the `examples/esp32` files that deal with `TRng` (truly rng) were not quite up-to-date with upstream API change. Modified according to the [migration guide](https://github.com/esp-rs/esp-hal/blob/main/esp-hal/MIGRATING-1.0.0-rc.0.md#rng-changes).

A question of opinion regarding the `trng_source` variable's name:

The guide uses `trng_source` - which "needs to be alive" for the `try_new()` to succeed. If I leave the name like that, there's a build warning (yikes). I also tried just an underscore (`let _ = ...`), but that seems to affect the lifespan of the value, which was.. surprising to me.

### DIFF
--- a/examples/esp32/src/bin/ble_bas_central_sec.rs
+++ b/examples/esp32/src/bin/ble_bas_central_sec.rs
@@ -3,6 +3,7 @@
 
 use embassy_executor::Spawner;
 use esp_hal::clock::CpuClock;
+use esp_hal::rng::{Trng, TrngSource};
 use esp_hal::timer::timg::TimerGroup;
 use esp_radio::Controller;
 use esp_radio::ble::controller::BleConnector;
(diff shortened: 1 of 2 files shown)


examples/esp32/src/bin/ble_bas_central_sec.rs
```diff
@@ -21,7 +22,8 @@ async fn main(_s: Spawner) {
     let timg0 = TimerGroup::new(peripherals.TIMG0);
     esp_preempt::start(timg0.timer0);
 
-    let mut rng = esp_hal::rng::Trng::new(peripherals.RNG, peripherals.ADC1);
+    let _trng_source = TrngSource::new(peripherals.RNG, peripherals.ADC1); // while alive, 'Trng::try_new()' succeeds
+    let mut trng = Trng::try_new().unwrap();
```
MabezDev: If you call this _after_ `esp_radio::init` you don't need the line above, using the rf subsystem will enable true rng :)

finalyards: That sounds splendid! 

..if it worked that way. By removing the first lines and moving second below `RADIO.init`, we get:

```
panicked at src/bin/ble_bas_peripheral_sec.rs:28:36:
called `Result::unwrap()` on an `Err` value: TrngSourceNotEnabled
```

I then looked into the `esp_radio` sources, and don't find a place where `TrngSource` is created. Also, such an entity needs to be "alive" (not dropped), for the `Trng::try_new()` to succeed.

I ended up simply moving the comment from the 1st lines to the 2nd. If people end up with a panic, it's great to have such a comment close by. I tried a few other ways too, and particularly dislike the term "alive" (which esp-hal uses, but...). The technically correct term might be "has not been dropped". Now using "accessible" - is that appropriate?   "Enabled" also used somewhere sounds like it just needs to be switched on and can be forgotten - that is not the case. It needs to be ... alive.

PR #456, opened 2025-09-23
suggestion on reducing deps in 'examples/apps'

finalyards:
I was collecting dependencies needed to use `trouble` in my project, and noticed some of the ones in `examples/apps/Cargo.toml` aren't really necessary, are they?

Contextual comments on each change.

### DIFF
--- a/examples/apps/Cargo.toml
+++ b/examples/apps/Cargo.toml
@@ -8,15 +8,16 @@ license = "MIT OR Apache-2.0"
 trouble-host = { path = "../../host", features = ["derive", "scan"] }
 bt-hci = { version = "0.5" }
 embassy-futures = "0.1.1"
-embassy-executor = { version = "0.9" }
 embassy-sync = { version = "0.7" }
 embassy-time = "0.5"
-embedded-hal = "1.0"
+#embedded-hal = "1.0"
 embedded-hal-async = "1.0"
 static_cell = "2"
-embedded-io = "0.6"
+#embedded-io = "0.6"
 heapless = "0.9"
 rand_core = { version = "0.6", default-features = false }
+
+# Used by 'security' feature
 embedded-storage-async = { version = "0.4.1", optional = true }
 sequential-storage = { version = "5.0.0", optional = true }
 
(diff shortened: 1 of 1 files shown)


examples/apps/Cargo.toml
```diff
@@ -8,15 +8,16 @@ license = "MIT OR Apache-2.0"
 trouble-host = { path = "../../host", features = ["derive", "scan"] }
 bt-hci = { version = "0.5" }
 embassy-futures = "0.1.1"
-embassy-executor = { version = "0.9" }
```
finalyards, line 11: Executor is clearly the domain of the main app, and those are in the platform specific folders. I would leave this out from here.

examples/apps/Cargo.toml
```diff
@@ -8,15 +8,16 @@ license = "MIT OR Apache-2.0"
...
 embassy-sync = { version = "0.7" }
 embassy-time = "0.5"
-embedded-hal = "1.0"
+#embedded-hal = "1.0"
```
finalyards, line 13: Not sure whether it can go. ESP32 samples compile without, and there are no references to `embedded_hal` (only `embedded_hal_async`) in the sources. But.. I'm not familiar with `embedded_hal_*` so please let me know, can it go. If so, I'd remove the line.

examples/apps/Cargo.toml
```diff
@@ -8,15 +8,16 @@ license = "MIT OR Apache-2.0"
...
 embedded-hal-async = "1.0"
 static_cell = "2"
-embedded-io = "0.6"
+#embedded-io = "0.6"
```
finalyards, line 16: Similar to `embedded_hal`:

not sure, haven't used myself. References only to `embedded_io_adapters` (in `serial-hci` and `tests`). Verdict = go or remain?

examples/apps/Cargo.toml
```diff
@@ -8,15 +8,16 @@ license = "MIT OR Apache-2.0"
...
 heapless = "0.9"
 rand_core = { version = "0.6", default-features = false }
+
+# Used by 'security' feature
```
finalyards, line 20: Would like such a comment, makes it more pronounced when one should have those in their own project.

PR #458, opened 2025-09-23
Comment change; relationship of 'default-packet-pool' and '...-mtu-'

finalyards:
>"[...] default-packet-pool-mtu and defeault-packet-pool-mtu."

That likely wants to be "default-packet-pool-size and default-packet-pool-mtu."?  Which one would you prefer to be mentioned first (which one is more important)?

Also, may I split the line to two lines? Such tails never get attention, which may have contrinuted to the mistakes.

---

More vitally, I stumbled on this when my bringing in of `trouble` didn't succeed. Tried this (from my `Cargo.toml`):

```toml
trouble-host    = { version = "0.3.0", default-features = false, features = [
    "default-packet-pool-mtu-255",
    "derive",
    "gatt",
    "peripheral",
    "security"
] }
```

Error message:

```
DEFMT_LOG=esp_hal=info,debug cargo build --release --features=defmt --example custom-emb
   Compiling ble-custom v0.0.0 (/home/ubuntu/ZOO.comms/comms/ble-custom)
error[E0412]: cannot find type `DefaultPacketPool` in module `trouble_host::prelude`
   --> examples/custom-emb/server_ble.rs:21:1
    |
 21 | #[gatt_server]
    | ^^^^^^^^^^^^^^ not found in `trouble_host::prelude`
    |
```

If we want to keep `default-packet-pool` and its default-steering fellows separate, I could steer this PR to making a more informative error message (that one should enable both `...-mtu-...` and `default-packet-pool`. But couldn't we simply do it so that if one defines a default `mtu` or `size`, that automatically implies `default-packet-pool`, as well.

That would have made my initial - naive - attempt to just work.

I see no down side on that. The change is simply some implied features in the `trouble-host` `Cargo.toml`. 

My code stumbled on this, because I wanted to avoid the `central` feature from creeping in. I think that's a pretty common use case: many projects are either `peripheral` or `central`.

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -81,7 +81,7 @@ security = [ "dep:p256", "dep:aes", "dep:cmac", "dep:rand_chacha", "gatt", "dep:
 # generator (CSPRNG) to require a cryptographically secure seed
 dev-disable-csprng-seed-requirement = []
 
-# Default packet pool. Enabling this will make available a packet pool tuned according to the default-packet-pool->>mtu<< and default-packet-pool->>mtu<<.
+# Enabling this will make available a packet pool tuned according to the default-packet-pool-mtu and default-packet-pool-size.
 default-packet-pool = []
 
 # Optimization where l2cap SDU reassembly saves some buffer copy.


finalyards:
I would rather suggest this kind of change:

Currently: 

```
default = ["peripheral", "central", "gatt", "derive", "default-packet-pool"]
```

I see that three features, `peripheral`, `central` and `scan` are similar, describing the functionality one would like to have from the library. Of these, `scan` extends `central` (and could thus imply it, perhaps). Would it be okay to drop all of these from the defaults?  To me, explicit is better than implicit, and this way I could use the default features, just add `peripheral` on top.

For the examples, this causes more code, sure. But how common is it in real projects, to want to have both peripheral and central functionality?  I'm newbie to embedded BLE, so please correct me if I'm writing nonsense...

---

Finally. Would it be welcomed/supported that `build.rs` (which is already there) would check feature set validity? I do this in my own projects, and like it. It gives a fast fail if one has conflicting feature sets (e.g. `scan` without `central`).  If such a check is welcomed, I can craft the PR based on documented relations of the features - and then we can bake it further.

lulf:
> I would rather suggest this kind of change:
> 
> Currently:
> 
> ```
> default = ["peripheral", "central", "gatt", "derive", "default-packet-pool"]
> ```
> 
> I see that three features, `peripheral`, `central` and `scan` are similar, describing the functionality one would like to have from the library. Of these, `scan` extends `central` (and could thus imply it, perhaps). Would it be okay to drop all of these from the defaults? To me, explicit is better than implicit, and this way I could use the default features, just add `peripheral` on top.
> 
> For the examples, this causes more code, sure. But how common is it in real projects, to want to have both peripheral and central functionality? I'm newbie to embedded BLE, so please correct me if I'm writing nonsense...
> 

This is quite common if you're building a mesh network for instance. I'd like to keep the default features, but I'll change the gen_features.py script to add the dependency so that packet pool gets enabled if you configure the mtu or size.

> Finally. Would it be welcomed/supported that `build.rs` (which is already there) would check feature set validity? I do this in my own projects, and like it. It gives a fast fail if one has conflicting feature sets (e.g. `scan` without `central`). If such a check is welcomed, I can craft the PR based on documented relations of the features - and then we can bake it further.

That would be much appreciated!

finalyards:
> > Finally. Would it be welcomed/supported that `build.rs` (which is already there) would check feature set validity? I do this in my own projects, and like it. It gives a fast fail if one has conflicting feature sets (e.g. `scan` without `central`). If such a check is welcomed, I can craft the PR based on documented relations of the features - and then we can bake it further.
> 
> That would be much appreciated!

Great! :) I'll start that next week, and make a separate PR. Will now tune this so that the one-line-comment-change can be separately merged.

finalyards:
@lulf ready for merge

lulf:
Hold on... some additional changes incoming on main

lulf:
Please rebase again

lulf:
Nevermind, it's just a doc change :)

PR #463, opened 2025-09-29
Changes to build.rs files

finalyards:
As discussed [here](https://github.com/embassy-rs/trouble/discussions/462).

### DIFF
--- a/benchmarks/nrf-sdc/build.rs
+++ b/benchmarks/nrf-sdc/build.rs
@@ -32,15 +32,15 @@ fn main() {
         .unwrap()
         .write_all(linker_data())
         .unwrap();
-    println!("cargo:rustc-link-search={}", out.display());
+    println!("cargo::rustc-link-search={}", out.display());
 
     // By default, Cargo will re-run a build script whenever
     // any file in the project changes. By specifying `memory.x`
     // here, we ensure the build script is only re-run when
     // `memory.x` is changed.
-    println!("cargo:rerun-if-changed=memory.x");
+    println!("cargo::rerun-if-changed=memory.x");
 
-    println!("cargo:rustc-link-arg-bins=--nmagic");
-    println!("cargo:rustc-link-arg-bins=-Tlink.x");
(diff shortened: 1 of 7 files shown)


finalyards:
The `test` config was indeed needed, by CI.

Took also the `cargo::rerun-if-changed=build.rs` back. I feel I now have a better grasp of what it's about. Running the build script (`build.rs`) is not the same as building the crate. The line informs Cargo that "the build script only needs rerunning if...". 

This doesn't affect rebuilds. If you change the code, Cargo will build the project. But the *build script* would only run, when actually needed.

I changed the comments to reflect this (please correct if I'm wrong):

```
    // We don't use any external files: only run the build script if it has changed.
    // Otherwise, Cargo will re-run it on each build.
```

The build script *always* gets re-run if it has changed. Cargo would not need us to confirm that, but it's the suggested way to say: hey, I don't need to be bothered when files around me change.

One more:

Should I format the lines, with some command, before a PR?   I noticed that many comments (like this) are way shorter (line wrap happens earlier). Let me knoww.

lulf:
> The `test` config was indeed needed, by CI.
> 
> Took also the `cargo::rerun-if-changed=build.rs` back. I feel I now have a better grasp of what it's about. Running the build script (`build.rs`) is not the same as building the crate. The line informs Cargo that "the build script only needs rerunning if...".
> 
> This doesn't affect rebuilds. If you change the code, Cargo will build the project. But the _build script_ would only run, when actually needed.
> 
> I changed the comments to reflect this (please correct if I'm wrong):
> 
> ```
>     // We don't use any external files: only run the build script if it has changed.
>     // Otherwise, Cargo will re-run it on each build.
> ```
> 
> The build script _always_ gets re-run if it has changed. Cargo would not need us to confirm that, but it's the suggested way to say: hey, I don't need to be bothered when files around me change.
> 
> One more:
> 
> Should I format the lines, with some command, before a PR? I noticed that many comments (like this) are way shorter (line wrap happens earlier). Let me knoww.

Use `cargo +nightly fmt`, because the rustfmt.toml has some nightly features.

PR #464, opened 2025-09-29
Early check of 'host' feature consistency

finalyards:
As discussed, here's a start for checking feature invariants in `build.rs`. 

This is something I've taken to be a habit of doing in my own projects. It seems to help to *understand* the relations between features - if there are any. At the moment, these are documented, and some checked in the source (which is fine; they can well be checked in two different levels - it serves slightly different aims).

I don't mean that this needs to get in. It's just a sample on how this works.

If possible, I've used `compile_error!`, which provides better error messages than panicing when the script is actually run. I think also the remaining cases can be made to use `compile_error!` by using consts. Let me know if that's something worth doing.

In addition to the tests, there are some suggestions to the `Cargo.toml`: any feature (looking at `scan`) that only makes sense in relation to another (`central`) should/could imho infer that feature. Us developers often check the "fingerprint" of a package by reading its Cargo.toml so this is partly/mostly documentation.

I'm aware the `fault-packet-pool-{size|mtu]-{N}` are generated. @lulf mentioned he'd (separately) work on the Python script, to - perhaps - feed in the `"default-packet-pool"` in them.

Anyways, it's a start. I'm getting to know the source more fully, and this helped.

### DIFF
--- a/host/build.rs
+++ b/host/build.rs
@@ -33,6 +33,84 @@ fn main() {
     println!("cargo::rustc-check-cfg=cfg(test)");
     println!("cargo:rerun-if-changed=build.rs");
 
+    // Check feature usage.
+    //
+    // Helps both document inter-feature relations, and stop builds ASAP if invariants are not
+    // fulfilled. For the developer, this leads to more straightforward error messages.
+    //
+    // Reference: Cargo features > https://embassy.dev/trouble/#_cargo_features
+    {
+        // 'scan' needs 'central'
+        #[cfg(all(feature = "scan", not(feature = "central")))]
+        compile_error!("Feature 'scan' also needs 'central' to be enabled.");
+
+        // Also in 'fmt.rs'
+        #[cfg(all(feature = "defmt", feature = "log"))]
+        compile_error!("You may not enable both `defmt` and `log` features.");
+
+        // Also in 'lib.rs'
+        #[cfg(not(any(feature = "central", feature = "peripheral")))]
+        compile_error!("Must enable at least one of: `central`, `peripheral`");
+
+        //...
+
+        // Only one of (or none) 'default-packet-pool-size-{X}' is allowed
+        {
+            let n = 0;
+            #[cfg(feature = "default-packet-pool-size-1")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-2")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-4")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-8")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-16")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-32")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-64")]
+            let n = n+1;
+            #[cfg(feature = "default-packet-pool-size-128")]
+            let n = n+1;
+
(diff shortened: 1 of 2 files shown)


host/build.rs
```diff
@@ -33,6 +33,84 @@ fn main() {
...
+    // Check feature usage.
+    //
+    // Helps both document inter-feature relations, and stop builds ASAP if invariants are not
+    // fulfilled. For the developer, this leads to more straightforward error messages.
+    //
+    // Reference: Cargo features > https://embassy.dev/trouble/#_cargo_features
+    {
+        // 'scan' needs 'central'
+        #[cfg(all(feature = "scan", not(feature = "central")))]
+        compile_error!("Feature 'scan' also needs 'central' to be enabled.");
```
HaoboGu, line 45: This seems unnecessary since `scan = ["central"]` is added in `Cargo.toml`?

host/build.rs
```diff
@@ -33,6 +33,84 @@ fn main() {
...
+        // Also in 'lib.rs'
+        #[cfg(not(any(feature = "central", feature = "peripheral")))]
+        compile_error!("Must enable at least one of: `central`, `peripheral`");
+
+        //...
+
+        // Only one of (or none) 'default-packet-pool-size-{X}' is allowed
+        {
+            let n = 0;
+            #[cfg(feature = "default-packet-pool-size-1")]
```
HaoboGu, line 60: This is already checked in [build.rs](https://github.com/embassy-rs/trouble/blob/a143dcdd3fd9f4e843eafa3f568c2fa89650687e/host/build.rs#L80)

You will get

```
  --- stderr

  thread 'main' panicked at /Users/haobogu/.cargo/registry/src/rsproxy.cn-e3de039b2554c837/trouble-host-0.4.0/build.rs:80:29:
  multiple values set for feature DEFAULT_PACKET_POOL_MTU: 512 and 64
warning: build failed, waiting for other jobs to finish...
```

if you set multiple `default-packet-pool-size/mtu`

host/build.rs
```diff
@@ -33,6 +33,84 @@ fn main() {
...
+            #[cfg(not(feature = "default-packet-pool"))]
+            if n>0 {
+                panic!("📍 'default-packet-pool-size-{{X}}' feature also needs 'default-packet-pool' to be enabled.");
+            }
+        }
+
+        // Only one of (or none) 'default-packet-pool-mtu-{X}' is allowed
+        {
+            let n = 0;
+            #[cfg(feature = "default-packet-pool-mtu-27")]
```
HaoboGu, line 91: Checked in build.rs too

host/build.rs
finalyards: It is. But it's also intentional: I see there to be three layers to features: 

- Cargo.toml provides them
- build.rs can check them
- source code uses them

In my projects, I've found it nice to have a one place (build.rs) that collects the feature-interdependencies together, and just checks the invariants. I normally forget about this part, soon, because it doesn't need to be changed. It does remind you if you do changes elsewhere - it's merely a safeguard.

Also, sources can (and should, imho) check feature relations, again. It does no harm. But that check is *local*, it's only checking the condition relevant to that piece of code.

Whether any of this is needed in the `trouble` project is up for the discussion that this PR is helping to raise.

host/build.rs
finalyards: Oh, I indeed missed that!  Good point.

finalyards:
@lulf I have no idea what causes the CI to fail. Tried a modified `ci.sh` locally, and it passes. Could you relaunch the CI run?

Issue #423, opened 2025-07-15
Update `rand_core` to 0.9


HaoboGu:
I found that both `trouble-host` and examples use `rand_core = 0.6`. Is there any particular reason to not bump version to latest?

AndWass:
I made a quick attempt but one of the dependencies of p256 is stuck on rand_core 0.6.4 so you need that version to get the correct trait versions.

lulf:
Yeah, would be great to update it, but I believe the p256 dependency is where I got stuck as well.

lure23:
Their [version 0.14 (unreleased) CHANGELOG](https://github.com/RustCrypto/elliptic-curves/blob/86e55d0a94fe7aa531356823e393a6d6f08deda1/sm2/CHANGELOG.md?plain=1#L14) lists: 

>Update to `rand_core` v0.9 ([diff](https://github.com/RustCrypto/elliptic-curves/pull/1125/files))

So the issue becomes pushing for a crate release, [New crate versions?](https://github.com/RustCrypto/elliptic-curves/issues/1068)

---

...and the *actual* tracking issue is: [Tracking issue for next release series](https://github.com/RustCrypto/traits/issues/1571)

finalyards:
For those interested, the [`pr-rand-catchup` branch](https://github.com/finalyards/trouble.fork/tree/pr-rand-catchup) is now up to the latest versions. 

- `rand` is preparing for 0.10 release; I jumped over 0.9 and preparing for it
- all dependencies are Release Candidate status (no `pre`)
- there are some API changes where I'd like to have a discussion with people better informed with `trouble` - but this can also wait until, say, January

🧚‍♀️ Usage of the branch would be welcome, so that we can see if I've messed up something. I only have ESP32 (C3, C6) and nRF54 hardware available.

mawoka-myblock:
Any updates?

PR #474, opened 2025-10-06
Adapted for latest `rand` APIs

finalyards:
The version of `rand` used in the project has fallen behind, mainly due to `p256` still relying on the earlier version.

I wanted to give it a try, and attached is a PR that:

- raises `rand` to `0.10.0-rc.0` (currently latest)
- cleans up some dependencies that seem to be not used
- fixes some conditional compilation, especially with regard to `rand` (`security` feature)
- occasional non-functional cleanup..

The [changes to 0.9.x were extensive](https://github.com/rust-random/rand/blob/master/CHANGELOG.md), covering not only the API but the whole `rand` ecosystem, as well.

I chose to push past the 0.9, since we already can see what the 0.10 API would look like. Happy to keep maintaining this, as a PR, until the 0.10 proper is released.

Some questions to authors in the comments below. I need guidance.

Discussed previously: https://github.com/embassy-rs/trouble/issues/423

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -72,7 +72,9 @@ controller-host-flow-control = []
 connection-metrics = []
 # Enable additional channel metrics
 channel-metrics = []
-security = ["dep:aes", "dep:cmac", "dep:p256", "dep:rand", "gatt"]
+security = [ "dep:p256", "dep:aes", "dep:cmac", "dep:rand", "gatt" ]
+# Enable LE Legacy Pairing (BLE 4.0/4.1 fallback when peer doesn't support Secure Connections)
+legacy-pairing = ["security"]
 # For development. Disable security manager cryptographically secure pseudorandom number
 # generator (CSPRNG) to require a cryptographically secure seed
 dev-disable-csprng-seed-requirement = []
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -8,6 +8,7 @@
 use core::mem::MaybeUninit;
 
 use advertise::AdvertisementDataError;
+use bt_hci::cmd::le::LeReadMinimumSupportedConnectionInterval;
 use bt_hci::cmd::status::ReadRssi;
 use bt_hci::cmd::{AsyncCmd, SyncCmd};
 use bt_hci::param::{AddrKind, BdAddr};
@@ -23,7 +24,7 @@ use crate::channel_manager::ChannelStorage;
 use crate::connection::Connection;
 use crate::connection_manager::ConnectionStorage;
 #[cfg(feature = "security")]
-pub use crate::security_manager::{BondInformation, IdentityResolvingKey, LongTermKey};
+pub use crate::security_manager::{BondInformation, IdentityResolvingKey, LongTermKey, Reason as PairingFailedReason};
 pub use crate::types::capabilities::IoCapabilities;
 
 /// Number of bonding information stored
@@ -92,7 +93,7 @@ pub mod prelude {
     pub use crate::attribute_server::*;
     #[cfg(feature = "central")]
     pub use crate::central::*;
-    pub use crate::connection::*;
+    pub use crate::connection::{ConnectRateParams, *};
     #[cfg(feature = "gatt")]
     pub use crate::gap::*;
     #[cfg(feature = "gatt")]
@@ -107,7 +108,9 @@ pub mod prelude {
     #[cfg(feature = "scan")]
     pub use crate::scan::*;
     #[cfg(feature = "security")]
-    pub use crate::security_manager::{BondInformation, IdentityResolvingKey, LongTermKey};
+    pub use crate::security_manager::{
+        BondInformation, IdentityResolvingKey, LongTermKey, Reason as PairingFailedReason,
+    };
     pub use crate::types::capabilities::IoCapabilities;
     #[cfg(feature = "gatt")]
     pub use crate::types::gatt_traits::{AsGatt, FixedGattValue, FromGatt};
@@ -268,7 +271,7 @@ pub enum Error {
     Att(AttErrorCode),
     #[cfg(feature = "security")]
     /// Error from the security manager
-    Security(crate::security_manager::Reason),
+    Security(PairingFailedReason),
     /// Insufficient space in the buffer.
     InsufficientSpace,
     /// Invalid value.
@@ -633,6 +636,20 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
             .set_io_capabilities(io_capabilities);
     }
 
+    /// Enable or disable secure connections only mode.
+    ///
+    /// When enabled, legacy pairing is rejected even if the `legacy-pairing` feature is compiled in.
+    /// This matches the BLE spec's "Secure Connections Only Mode" (Vol 3, Part C, Section 10.2.4).
+    ///
+    /// Only relevant if the feature `legacy-pairing` is enabled.
+    #[cfg(feature = "legacy-pairing")]
+    pub fn set_secure_connections_only(&self, enabled: bool) {
+        self.host
+            .connections
+            .security_manager
+            .set_secure_connections_only(enabled);
+    }
+
     /// Build the stack.
     pub fn build(&'stack self) -> Host<'stack, C, P> {
         #[cfg(all(feature = "security", not(feature = "dev-disable-csprng-seed-requirement")))]
@@ -668,6 +685,16 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
         C: ControllerCmdAsync<T>,
     {
         self.host.async_command(cmd).await
+    }
+
+    /// Read the minimum supported connection interval from the controller.
+    pub async fn read_minimum_supported_connection_interval(
+        &self,
+    ) -> Result<<LeReadMinimumSupportedConnectionInterval as SyncCmd>::Return, BleHostError<C::Error>>
+    where
+        C: ControllerCmdSync<LeReadMinimumSupportedConnectionInterval>,
+    {
+        self.host.command(LeReadMinimumSupportedConnectionInterval::new()).await
     }
 
     /// Read current host metrics
--- a/host/src/security_manager/pairing/central.rs
+++ b/host/src/security_manager/pairing/central.rs
@@ -7,7 +7,6 @@ use rand::RngExt; // brings in 'Rng::sample()'
 
 use crate::codec::{Decode, Encode};
 use crate::connection::{ConnectionEvent, SecurityLevel};
-use crate::security_manager::constants::ENCRYPTION_KEY_SIZE_128_BITS;
 use crate::security_manager::crypto::{Confirm, DHKey, MacKey, Nonce, PublicKey, SecretKey};
 use crate::security_manager::pairing::util::{
     choose_pairing_method, make_confirm_packet, make_dhkey_check_packet, make_pairing_random, make_public_key_packet,
@@ -147,6 +146,7 @@ struct PairingData {
     ltk: Option<LongTermKey>,
     timeout_at: Instant,
     bond_information: Option<BondInformation>,
+    user_initiated: bool,
 }
 
 impl PairingData {
@@ -162,6 +162,15 @@ pub struct Pairing {
 }
 
 impl Pairing {
+    pub fn result(&self) -> Option<Result<(), Error>> {
+        let step = self.current_step.borrow();
+        match step.deref() {
+            Step::Success => Some(Ok(())),
+            Step::Error(e) => Some(Err(e.clone())),
+            _ => None,
+        }
+    }
+
     pub fn timeout_at(&self) -> Instant {
         let step = self.current_step.borrow();
         if matches!(step.deref(), Step::Success | Step::Error(_)) {
@@ -207,6 +216,7 @@ impl Pairing {
             private_key: None,
             timeout_at: Instant::now() + crate::security_manager::constants::TIMEOUT_DISABLE,
             bond_information: None,
+            user_initiated: false,
         };
         Self {
             pairing_data: RefCell::new(pairing_data),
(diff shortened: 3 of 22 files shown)


ci.sh
```diff
@@ -48,3 +48,5 @@ cargo clippy --manifest-path ./host/Cargo.toml --features gatt,peripheral,centra
 cargo test --manifest-path ./host/Cargo.toml --lib -- --nocapture
 cargo test --manifest-path ./host/Cargo.toml --no-run -- --nocapture
 cargo test --manifest-path ./examples/tests/Cargo.toml --no-run -- --nocapture
+
+cargo test --manifest-path ./host/Cargo.toml --features security  # Q: what place above should this be fitted in?  Should '--lib' or '--no-run' be used??
```
finalyards: >Edit: note that this snippet is now "outdated" - but the discussion is valid.

I'd like to understand the intention behind these two lines:

```
cargo test --manifest-path ./host/Cargo.toml --lib -- --nocapture
cargo test --manifest-path ./host/Cargo.toml --no-run -- --nocapture
```

The problem I detected was that, since `--feature=security` was not involved, whole source files did not end up compiled. Adding the same two lines with `--feature security` was my initial attempt, but when I looked at the parameters..

`--lib` compiles and runs the tests, within the source files.
`-- --nocapture` places their stdout output on the console (fine)

`--no-run` compiles *all areas* (including `tests/**.rs`, presumably), but does not run

>Since it doesn't run anything, `-- --nocapture` is unused on that command.

To remove the overlap, I would use:

```
cargo test --manifest-path ./host/Cargo.toml --lib -- --nocapture
cargo test --manifest-path ./host/Cargo.toml --tests --no-run
```

Is this okay?   Would introduce such changes in a separate PR, just add duplicated lines with `--feature security` here, because it does the trick!

examples/apps/Cargo.toml
```diff
@@ -8,14 +8,12 @@ license = "MIT OR Apache-2.0"
 trouble-host = { path = "../../host", features = ["derive", "scan"] }
 bt-hci = { version = "0.6" }
 embassy-futures = "0.1.1"
-embassy-sync = { version = "0.7" }
+embassy-sync = { version = "0.7" }  # host-macros
```
finalyards: marked the reason of existence for certain dependencies (here and `static_cell`, `heapless`, below).

examples/apps/Cargo.toml
```diff
@@ -8,14 +8,12 @@ license = "MIT OR Apache-2.0"
...
 embassy-time = "0.5"
-#embedded-hal = "1.0"
```
finalyards: such comments (only in this file) were left-overs from an earlier PR (and were meant to be deleted). Now are. :)

examples/apps/src/lib.rs
```diff
@@ -7,8 +7,10 @@ pub(crate) mod fmt;
 pub mod ble_advertise;
 pub mod ble_advertise_multiple;
 pub mod ble_bas_central;
+#[cfg(feature = "security")]
```
finalyards, line 11: `*_sec` were treated as feature-flagged, except here.

examples/nrf52/Cargo.toml
```diff
@@ -8,26 +8,20 @@ resolver = "2"
 embassy-executor = { version = "0.9", default-features = false, features = ["arch-cortex-m", "executor-thread", "defmt", "executor-interrupt"] }
 embassy-time = { version = "0.5", default-features = false, features = ["defmt", "defmt-timestamp-uptime"] }
 embassy-nrf = { version = "0.7", default-features = false, features = ["defmt", "time-driver-rtc1", "gpiote", "unstable-pac", "rt"] }
-embassy-futures = "0.1.1"
-embassy-sync = { version = "0.7", features = ["defmt"] }
 
-futures = { version = "0.3", default-features = false, features = ["async-await"]}
 nrf-sdc = { version = "0.4", default-features = false, features = ["defmt", "peripheral", "central"] }
 nrf-mpsl = { version = "0.3", default-features = false, features = ["defmt", "critical-section-impl"] }
-bt-hci = { version = "0.6", default-features = false, features = ["defmt"] }
```
finalyards: Five deps that seem unneeded: the `ci.sh` runs if they are gone.

I'm *not* fully sure, though, since don't have nRF hardware myself.

- [x] please confirm

examples/nrf52/Cargo.toml
```diff
@@ -8,26 +8,20 @@ resolver = "2"
...
 trouble-host = { path = "../../host", features = ["derive", "scan"] }
 trouble-example-apps = { version = "0.1.0", path = "../apps", features = ["defmt"] }
 
 defmt = "1.0"
 defmt-rtt = "1.0"
 panic-probe = { version = "1.0", features = ["print-defmt"] }
 
-cortex-m = { version = "0.7.6" }
+chacha20 = { version = "0.10.0-rc.2", default-features = false, features = ["rng"], optional = true }
```
finalyards: `rand` 0.10 changelog:

>The dependency on `rand_chacha` has been replaced with a dependency on `chacha20`.

Moved it upwards, to have the list in roughly alphabetical order (imho, a good idea). Do you agree with this, or should it be in-place of the earlier `rand_core`?

examples/nrf52/Cargo.toml
```diff
@@ -8,26 +8,20 @@ resolver = "2"
...
 cortex-m-rt = "0.7.0"
-rand = { version = "0.8.5", default-features = false }
+rand_core = { version = "0.9.3", default-features = false, optional = true }
```
finalyards: `rand_core` is interesting. [crates.io for 0.9.3](https://docs.rs/rand_core/latest/rand_core/) states:

>This crate is mainly of interest to crates publishing implementations of [RngCore](https://docs.rs/rand_core/latest/rand_core/trait.RngCore.html). Other users are encouraged to use the [rand](https://docs.rs/rand) crate instead which re-exports the main traits and error types.

But the current [README](https://github.com/rust-random/rand/blob/master/rand_core/README.md) says it differently:

>The main [rand](https://crates.io/crates/rand) crate re-exports most items defined in this crate, along with tools [...] Most users should prefer to use the main [rand](https://crates.io/crates/rand) crate.

I feel there's a justification for keeping `rand_core`, if the code only needs the *interface* of the RNG's, and doesn't actually create random numbers. What do you think?

That's the approach I took here - `nrf-sdc` - but in `host` I've opted for `rand` and use the interfaces as `rand::rand_core::...`.

examples/rp-pico-2-w/Cargo.toml
```diff
@@ -48,7 +47,6 @@ debug = 2
 #embassy-hal-internal = {path = "../../../embassy/embassy-hal-internal"}
 #nrf-sdc = { path = "../../../nrf-sdc/nrf-sdc" }
 #nrf-mpsl = { path = "../../../nrf-sdc/nrf-mpsl" }
-#bt-hci = { path = "../../../bt-hci" }
```
finalyards, line 51: can the line go? `bt-hci` not needed in `rp-pico-2-w`

host/Cargo.toml
```diff
@@ -28,10 +29,8 @@ embassy-futures = "0.1"
 futures = { version = "0.3", default-features = false }
 heapless = "0.9"
 trouble-host-macros = { path = "../host-macros", version = "0.3.0", optional = true }
-p256 = { version = "0.13.2", default-features = false, features = ["ecdh","arithmetic"], optional = true }
-rand_core = "0.6"
-rand_chacha = { version = "0.3", default-features = false, optional = true }
-rand = { version="0.8", default-features = false, optional = true}
+p256 = { version = "0.14.0-pre.11", default-features = false, features = ["ecdh","arithmetic"], optional = true }
```
finalyards: The `0.14.0-pre.3` [seems to be the release](https://github.com/RustCrypto/elliptic-curves/blob/p256/v0.14.0-pre.3/p256/Cargo.toml#L39) that has the `rand` 0.6 -> 0.9 upgrade.

This actually is the **BIG THING** about the PR. Whether we want to jump onto `-pre` or wait until `p256` 0.14.0 proper has been released.

host/Cargo.toml
```diff
@@ -50,10 +49,9 @@ embedded-io = { version = "0.6.1" }
 tokio-serial = "5.4"
 env_logger = "0.11"
 critical-section = { version = "1", features = ["std"] }  # needed for CI builds
-rand = "0.8.5"
-rand_core = { version = "0.6", features = ["getrandom"]}
+rand = { version = "0.10.0-rc.0", default-features = false, features = ["os_rng","thread_rng"] }
 heapless = "0.9"
-embassy-executor = { version = "0.9", features = ["arch-std", "executor-thread"]}
+#embassy-executor = { version = "0.9", features = ["arch-std", "executor-thread"]}
```
finalyards: that... line... can go, right? :)

host/src/lib.rs
```diff
@@ -15,7 +15,8 @@ use bt_hci::FromHciBytesError;
 use embassy_time::Duration;
 #[cfg(feature = "security")]
 use heapless::Vec;
-use rand_core::{CryptoRng, RngCore};
+#[cfg(feature = "security")]
+use rand::rand_core::{CryptoRng, RngCore};
```
finalyards: Note: here, we could go on with the earlier lines, if `rand_core` is added (alongside `rand`), as a dependency to `trouble-host`. It's a compromise between either:

- shorter code; two dependencies
- slightly longer code; just `rand` dep

The upstream seems to recomment going `rand` only, but it's really our choice. Which way is better?

- [ ] please inform of the choice.

host/src/lib.rs
```diff
@@ -436,14 +437,14 @@ pub trait Controller:
     + ControllerCmdAsync<LeCreateConn>
     + ControllerCmdSync<LeClearFilterAcceptList>
     + ControllerCmdSync<LeAddDeviceToFilterAcceptList>
-    + for<'t> ControllerCmdSync<LeSetAdvEnable>
+    + ControllerCmdSync<LeSetAdvEnable>
```
finalyards: unrelated cleanups - I cannot see why `for<'t>` would matter, there

host/src/lib.rs
```diff
@@ -562,7 +563,7 @@ pub fn new<
     }
 
     // Safety:
-    // - HostResources has the exceeding lifetime as the returned Stack.
+    // - HostResources has the exceeding lifetime as the returned Stack.    <<--??
```
finalyards: I marked this, because had difficulty (still have) understanding that comment. Feel free to educate me here; I'll remove the mark in the next revise.

host/src/lib.rs
```diff
@@ -620,11 +624,11 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
         }
         self
     }
+    #[cfg(feature = "security")]
     /// Set the IO capabilities used by the security manager.
     ///
     /// Only relevant if the feature `security` is enabled.
     pub fn set_io_capabilities(self, io_capabilities: IoCapabilities) -> Self {
-        #[cfg(feature = "security")]
```
finalyards, line 631: Moved the conditional above the method name. This was made possible by making the `security` feature handling more consistent, elsewhere.

Should be beneficial: one only needs to see such method signatures in IDE if the `security` feature is enabled.

host/src/security_manager/pairing/central.rs
```diff
@@ -386,8 +386,13 @@ impl Pairing {
...
                         PairingMethod::PassKeyEntry { central, .. } => {
                             if central == PassKeyEntryAction::Display {
+                                // Note: 'Uniform::new*()' are aimed at cases where multiple samples would be fetched.
+                                //      "Constructors may do extra work up front to allow faster sampling of multiple values.
+                                //       Where only a single sample is required it is suggested to use Rng::random_range [...]"
+                                //pairing_data.local_secret_ra = rng.random_range(0..=999999);
                                 pairing_data.local_secret_ra =
-                                    rng.sample(rand::distributions::Uniform::new_inclusive(0, 999999));
+                                    rng.sample(rand::distr::Uniform::new_inclusive(0, 999999).unwrap());
+
```
finalyards: Here, I chose to leave the original code (after API catch-up), but the `rand` authors recommend:

>Constructors may do extra work up front to allow faster sampling of multiple values. Where only a single sample is required it is suggested to use `Rng::random_range [...]"`.

Should we?

finalyards:
/test

@lulf Would you, the tests should pass now that #470 is in. thx

finalyards:
A bit about the status:

- it's ready for reviews, and use, actually
- when I get conclusion on the discussion places (might tag them with some eye catching emoji...), I'll do fine-tuning and
- ..squash into a single commit

finalyards:
Issues unrelated to random/crypto are now in https://github.com/embassy-rs/trouble/pull/476 . If we merge that in, I can remove them from this PR.

finalyards:
The CI does not really fail - [it says "No space left of device"](https://github.com/embassy-rs/trouble/actions/runs/18651869851/job/53171553648?pr=474#step:5:576).

 The PR now only includes randomness, and security-related changes (+ ~4 unrelated comments). While it introduces a dependency on `rand = "0.10.0-rc.0"`, `chacha20 = "0.10.0-rc.2"` and `p256 = "0.14.0-pre.11"`, I personally feel it would be okay to merge it in: Trouble itself is still in the `0.x` times and developing; the dependent crates are slow-moving but seem steady to me, at least API-wise.
 
Could this be considered for a merge, or do we need to wait (in which case I'll mark this back as a "draft" and bring it back when the dependencies have released solid releases).

finalyards:
NB. CI failing is due to the system not using the modified `.github` actions (in the PR), but the existing `main` ones. I'll come back to this, later... (about 1 week)

finalyards:
The CI failing is simply because the `rand` + crypto RC ecosystem is still in flux. I keep tracking the upstream (RC) versions and maintain this PR until they become proper releases. At that point, I'll of course clean up the PR and squash it into a single commit. Seems to be taking its time..

finalyards:
Status of `rand` dependencies:

- `rand_core` has reached 0.10 (proper)
- others are now catching up with this; it causes a domino effect
- shouldn't be more than some weeks, until we can build TrouBLE on `rand` 0.10, without RC's :)

Currently fails CI:

```
+ cargo check --release --manifest-path examples/esp32/Cargo.toml --target riscv32imc-unknown-none-elf --features esp32c3
    Updating crates.io index
error: failed to select a version for `rand_core`.
    ... required by package `rand v0.10.0-rc.9`
    ... which satisfies dependency `rand = "^0.10.0-rc.9"` of package `trouble-host v0.5.1 (/home/ubuntu/trouble.fork/host)`
    ... which satisfies path dependency `trouble-host` (locked to 0.5.1) of package `trouble-esp32-examples v0.1.0 (/home/ubuntu/trouble.fork/examples/esp32)`
versions that meet the requirements `^0.10.0-rc-6` (locked to 0.10.0-rc-6) are: 0.10.0-rc-6

all possible versions conflict with previously selected packages.

  previously selected package `rand_core v0.10.0`
    ... which satisfies dependency `rand_core = "^0.10.0"` of package `trouble-example-apps v0.1.0 (/home/ubuntu/trouble.fork/examples/apps)`
    ... which satisfies path dependency `trouble-example-apps` (locked to 0.1.0) of package `trouble-esp32-examples v0.1.0 (/home/ubuntu/trouble.fork/examples/esp32)`

failed to select a version for `rand_core` which could resolve this conflict
```

This is due to `rand` still not having been updated to `rand_core` 0.10. `elliptic-curves` has, so we just need to wait...

In the mean time, we *can* work on these:

- [ ] #521 

   Having this merged to `main` will reduce the footprint of this PR; I think examples are the only place where `rand_core` is still directly depended upon.

- [ ] Tests

   Changes in `chacha20` mean one can no longer seed `ChaCha12Rng` from an `u64`. My efforts to keep the public keys as-is failed, so I wish(!!) someone who knows about the tests more, to have a look.  At the moment, I've disabled two tests [here](https://github.com/finalyards/trouble.fork/blob/pr-rand-catchup/host/src/security_manager/pairing/peripheral.rs#L740) and [here](https://github.com/finalyards/trouble.fork/blob/pr-rand-catchup/host/src/security_manager/pairing/peripheral.rs#L889). cc @lulf @AndWass

PR #475, opened 2025-10-06
Update ci.yaml

finalyards:
CI seems broken; this changes the path to what exists in the repo.


finalyards:
Already here: https://github.com/embassy-rs/trouble/pull/470

PR #480, opened 2025-10-13
feat: add feature flag to conditionally support conn param request update

### LINKED ISSUE #479: ESP32 ble_bas_peripheral example discconects after a while
Running the `ble_bas_peripheral` of latest `main` ([2f8e936](https://github.com/embassy-rs/trouble/commit/2f8e936a2ebdd3373a01e31aff4b332b88bfc39d)) results in: 

```
trouble/examples/esp32 on  main is 📦 v0.1.0 via 🦀 v1.88.0 took 4s 
❯ cargo run --release --no-default-features --features=esp32c6 --target=riscv32imac-unknown-none-elf --bin ble_bas_peripheral
....
INFO - [custom_task] notifying connection of tick 23
INFO - [custom_task] RSSI: -44
INFO - [host] disconnection event on handle 0, reason: Connection Timeout
INFO - [gatt] disconnected: Err(Connection Timeout)
INFO - [adv] advertising
```

Tested in C3 and C6. 

This was discovered in latest esp-hal `main` ([4ee4d73](https://github.com/esp-rs/esp-hal/commit/4ee4d738531e7e5b7517fe8723e8f093233facd1)), where the `bas_peripheral` results in the same output. 

For reference, this issue does not appear in https://github.com/crimpdeq/crimpdeq-firmware/blob/main/Cargo.toml which uses `trouble-host@0.2.4` and `esp-hal@1.0.0-rc.0`&company

lulf:
Supporting this feature requires the user of trouble-host to accept the desired parameters.

Not tested, but should fix #479

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -28,6 +28,7 @@ cargo batch \
     --- build --release --manifest-path host/Cargo.toml --no-default-features --features gatt,peripheral,central,scan,controller-host-flow-control \
     --- build --release --manifest-path host/Cargo.toml --no-default-features --features gatt,peripheral,central,scan,controller-host-flow-control,connection-metrics,channel-metrics \
     --- build --release --manifest-path host/Cargo.toml --no-default-features --features gatt,peripheral,central,scan,controller-host-flow-control,connection-metrics,channel-metrics,l2cap-sdu-reassembly-optimization \
+    --- build --release --manifest-path host/Cargo.toml --no-default-features --features gatt,peripheral,central,scan,controller-host-flow-control,connection-metrics,channel-metrics,l2cap-sdu-reassembly-optimization,connection-params-update \
     --- build --release --manifest-path bt-hci-linux/Cargo.toml \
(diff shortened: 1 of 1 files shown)


HaoboGu:
When the feature is enabled, could "accept" be automatically called by the host? or it's users responsibility to choose accepting the conn param or disconnect?

lulf:
> > Not tested, but should fix #479
> 
> I can confirm that it fixes the issue, I tested both the `ble_bas_peripheral` example and my personal project and both worked as expected

Thanks for testing @SergioGasquez!

lulf:
> When the feature is enabled, could "accept" be automatically called by the host? or it's users responsibility to choose accepting the conn param or disconnect?

The challenge with that is that some of the parameters like min/max connection event length are not provided by the event received from the peer, so the host would have to select or remember these values for each connection. I think maybe longer term we can find a way to make it automatic, maybe with some kind of policy configuration that's part of the host configuration that would contain the 'defaults' for these parameters. 

I imagine that some applications want to be able to check a few things before allowing the connection parameters to be updated as well.

Issue #482, opened 2025-10-15
Policy on `Cargo.lock`s would be good


finalyards:
```
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   examples/esp32/Cargo.lock
	modified:   examples/nrf-sdc/Cargo.lock
```

That happened quite a bit, working on a PR today. Made me wonder, whether the lock files actually matter, for examples.

**Current situation**

`/.gitignore` [states](https://github.com/embassy-rs/trouble/blob/main/.gitignore#L3) that all `Cargo.lock`s should be ignored (not suggested to be placed to version control).

```
Cargo.lock
```

However, there clearly are instances already in the version control. Otherwise, I wouldn't have gotten the "modified" mentions, above.

```
./examples/apache-nimble/Cargo.lock
./examples/rp-pico-w/Cargo.lock
./examples/nrf-sdc/Cargo.lock
./examples/esp32/Cargo.lock
./examples/serial-hci/Cargo.lock
```

This is inconsistent, at best.

The `esp-hal` project has [this approach](https://github.com/esp-rs/esp-hal/blob/main/.gitignore#L6-L8) to including `Cargo.lock`s, within a repo:

```
# Remove Cargo.lock from gitignore if creating an executable, leave it for libraries
# More information here https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html
Cargo.lock
```

<strike>"Leave it for libraries". 

All the ones we have are for examples. I think they can go?    ((give a nudge and I'll make a PR that removes them, making the repo consistent with what's in its `/.gitignore`))

But shouldn't we actually *have* it for the libraries (`host` and `host-macros`)?</strike>

Should I add those in the same PR?

HaoboGu:
https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html is quite clear, `Cargo.lock` in examples(bin) can be kept, and for libs like `host` and `host-macros` they should not be committed.

lulf:
Yes, examples committed and libs not committed is the policy. We could move the .gitignore rules for ignoring lock files to the host and host-macros subdirectories.

finalyards:
Thanks!  I don't know why I got it exactly the other way round!

>We could move the .gitignore rules for ignoring lock files to the host and host-macros subdirectories.

I can craft such a PR.

[...]

PR #486, opened 2025-10-15
Cargo.lock handling to be same across examples

finalyards:
I'm... having second thoughts about this, but since I volunteered in #482 to make one, here goes.

We can continue discussion here - the PR is intended to fulfil the specs in that discussion.


finalyards:
I made a second branch, last week, the [diff](https://github.com/embassy-rs/trouble/compare/main...finalyards:trouble.fork:pr-Cargo-lock-stage2) to `origin/main` is there.

Disclaimer: I'm not claiming this to be the truth. I'm just trying to make the Cargo.lock policy least invasive for me, as a developer, and keep it useful for the repo. I first thought my approach is disruptive, but Copilot sees it more as a continuation of the stated Cargo guidelines. (granted, Copilot has a tendency to agree; you be the judge of it)

In short, I would:

- remove `Cargo.lock`'s from the `examples/`
- retain them for tests
- ..which involved adding one for `host/`
- ..with the assumption (mildly googled) that it will not be used by Cargo, when using `host` as a dependency. It's meant for the tests

This move keeps one or two `Cargo.lock`s so that CI caching can be based on them, and `cargo test --locked` would keep some consistency to the runs.

Let me know what you think. If you show 👍 for the approach, I'll make the PR represent it, instead.

ps. I asked Copilot to summarize the approach as the `CONTRIBUTING.md`.

finalyards:
The CI states [no space on device](https://github.com/embassy-rs/trouble/actions/runs/18694718765/job/53309087307?pr=486#step:5:609). 

I don't think I can fix that in any way? Let me know when there's a chance it might pass.

.gitignore
```diff
@@ -1,4 +1,16 @@
-**/target
-**/target_ci
+**/target/
+/target_ci/
```
lulf, line 2: This should still be **/target_ci  I think?

finalyards: There's only one `target_ci` used, ever, whereas `target` can be in the subdirs. I like to be explicit with these patterns - narrow is better than general.

finalyards:
@lulf I didn't intend to mislead, or hijack my own PR. I considered what you meant by "lgtm" and decided it was in the context of the [above comment](https://github.com/embassy-rs/trouble/pull/486#issuecomment-3421572009) - thus went ahead and replaced the original approach with that. 

This was a mistake. But since I still think that approach - no lock files needed for examples - is better, I will:

- restore this PR to where it was, and CLOSE IT.

I don't advocate the original change to the repo, any more. The branch I made in my fork remains available, if someone wants to use it as a reference. Sorry for the disturbance!!!

PR #487, opened 2025-10-15
stops unnecessary `.lock` regeneration 

finalyards:
The problem:

- `trouble-host` in `main` is now 0.5.0, but two lock files still have 0.4.0 in them.

This caused, at least for me, the IDE to insist that such locks must be updated. Yes, I could... but was working on PR's in the mean time so. Merging this PR should remove such a competition from the file system.

### DIFF
--- a/examples/esp32/Cargo.lock
+++ b/examples/esp32/Cargo.lock
@@ -1822,7 +1822,7 @@ dependencies = [
 
 [[package]]
 name = "trouble-host"
-version = "0.4.0"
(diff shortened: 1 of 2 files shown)


finalyards:
What does it mean that "example-tests" is failing?

lulf:
> What does it mean that "example-tests" is failing?

They were failing because the CI machine had some issues with connecting to the test hardware. Some of the examples are tested on real hardware.

PR #497, opened 2025-10-30
Fix for build.rs: double colon issue

jbinko:
# Fix incorrect Cargo directive in build.rs

## Description
The `build.rs` script currently uses a double colon (`::`) in the `cargo` directive, which is incorrect. Cargo expects a single colon (`:`) for key-value pairs in build script output.

## Current Behavior
```rust
println!("cargo::rustc-link-search={}", out.display());
```

## Expected Behavior
```rust
println!("cargo:rustc-link-search={}", out.display());
```

### DIFF
--- a/benchmarks/nrf-sdc/build.rs
+++ b/benchmarks/nrf-sdc/build.rs
@@ -32,15 +32,15 @@ fn main() {
         .unwrap()
         .write_all(linker_data())
         .unwrap();
-    println!("cargo::rustc-link-search={}", out.display());
+    println!("cargo:rustc-link-search={}", out.display());
 
     // By default, Cargo will re-run a build script whenever
     // any file in the project changes. By specifying `memory.x`
     // here, we ensure the build script is only re-run when
     // `memory.x` is changed.
-    println!("cargo::rerun-if-changed=memory.x");
+    println!("cargo:rerun-if-changed=memory.x");
 
-    println!("cargo::rustc-link-arg-bins=--nmagic");
-    println!("cargo::rustc-link-arg-bins=-Tlink.x");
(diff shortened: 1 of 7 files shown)


finalyards:
Does it, though?

I discussed this with Copilot, in September, and it - like you - kept firm that one colon is the *official* and two would be "ignored".

This brings a couple of questions to my mind:
- why do things still work?
- ..or do they?
- why hasn't CI picked up anything?

But sometimes AI gets confused. Especially when things have changed over time. 

Note that also [Cargo book](https://doc.rust-lang.org/cargo/reference/build-scripts.html) uses two colons:

```rust
println!("cargo::rerun-if-changed=src/hello.c");
```

I think I looked this back to the particular PR in Cargo that moved from one to two (keeping both equally functioning, for backwards compatibility). So, with the authority of the Book I think it's safe to say, a) it does not matter, b) two is more correct than one.

i.e. you've found an interesting spot, but ... the PR would be a step backwards, in time.

p.s. what we could do, is the reverse: add a clippy-like test that no such `println!` is with only one colon!! :)

finalyards:
```bash
#!/bin/bash
if grep -r --include="build.rs" -n 'println!("cargo:[^:]' .; then
    echo "Found 'build.rs' directives with single colon. Would you mind changing them to 'cargo::'."
    false
else
    echo "'build.rs' directives ok!"
fi
```

lulf:
I'd prefer to follow the cargo book and keep the existing double colon.

finalyards:
@jbinko Would you like to close this, or comment.

jbinko:
Will close this one.

PR #499, opened 2025-11-06
Fix lfclk cfg for nRF54 example

HaoboGu:
In nRF54 example, the ExternalXtal is used, the `mpsl_clock_lfclk_cfg_t` should use parameters for External Xtal as well. 

With this fix the ble_bas_peripheral example works fine for me


HaoboGu:
Tested nRF54L15 as BLE peripheral and BLE central + peripheral, works greeeeeeaaaaaaaaat

PR #503, opened 2025-11-24
chore: bump trouble-host version to v0.5.1

HaoboGu:


lulf:
(Had to force merge because the USB cable used for the example tests is taken right now :see_no_evil:)

PR #524, opened 2026-01-17
feat: Support static data for read-only characteristics

### LINKED ISSUE #357: `gatt_service` Proc Macro can't handle &'static str values
Hello you all were able to help me earlier with  #272 . I'm finally looking to pick that project back up and it seems that the  `gatt_service` macro can't handle `&'static str` values. I get the following error:
```
error[E0599]: no associated item named `Value` found for struct `DeviceInfoService` in the current scope
   --> rp_firmware/src/bluetooth.rs:110:63
    |
25  | #[gatt_service(uuid = "0000180a-0000-1000-8000-00805f9b34fb")]
    | -------------------------------------------------------------- associated item `Value` not found for this struct
...
110 |             AdStructure::ServiceUuids128(&[DeviceInfoService::Value]),
    |                                                               ^^^^^ associated item not found in `DeviceInfoService`

error[E0080]: could not evaluate static initializer
  --> rp_firmware/src/bluetooth.rs:38:20
   |
38 |     serial_number: &'static str,
   |                    ^ values of the type `[u8; usize::MAX]` are too big for the target architecture

Some errors have detailed explanations: E0080, E0599.
For more information about an error, try `rustc --explain E0080`.
```

Poking around a bit it looks like the result of the `AsGatt` implmentation of `&str` setting `MAX_SIZE = usize::MAX` :
```
impl AsGatt for &'static str {
    const MIN_SIZE: usize = 0;
    const MAX_SIZE: usize = usize::MAX;

    fn as_gatt(&self) -> &[u8] {
        self.as_bytes()
    }
}
```

and the macro having the following instantiation of `StaticCell`:
```
 static #name_screaming: static_cell::StaticCell<[u8; <#ty as trouble_host::types::gatt_traits::AsGatt>::MAX_SIZE]> = static_cell::StaticCell::new();
```

It looks like this would also impact the impl for `&[u8]` since that sets `MAX_SIZE` the same way

alexmoon:
This allows the gatt_service macro to use `&'static` values for read-only characteristics. It currently only supports references with an explicit `static` lifetime. This could probably be loosened since the compiler will complain if the value does not outlive the attribute table. It also only supports characteristics that only enable the `read` property.

To make this work properly, I needed to rework the `AsGatt`/`FromGatt`/`FixedSizeGatt` trait hierarchy slightly. There is now a blanket `AsGatt` impl for references to any type which impls `AsGatt`. I also loosened the type bounds on `Characteristic` to be `T: AsGatt + ?Sized` in most cases. This allows, for example, a `Characteristic<str>` to exist as the type of service characteristics with a `&'static str` value.

I added a `ble_dis_peripheral.rs` example demonstrating the feature.

Fixes #357

### DIFF
--- a/examples/apps/src/ble_dis_peripheral.rs
+++ b/examples/apps/src/ble_dis_peripheral.rs
@@ -1,0 +1,144 @@
+use embassy_futures::join::join;
+use trouble_host::prelude::*;
+
+/// Max number of connections
+const CONNECTIONS_MAX: usize = 1;
+
+/// Max number of L2CAP channels.
+const L2CAP_CHANNELS_MAX: usize = 2; // Signal + att
+
+// GATT Server definition
+#[gatt_server]
+struct Server {
+    info_service: DeviceInformationService,
+}
+
+/// Battery service
+#[gatt_service(uuid = service::DEVICE_INFORMATION)]
+struct DeviceInformationService {
+    /// Manufacturer name
+    #[characteristic(uuid = characteristic::MANUFACTURER_NAME_STRING, read, value = "Embassy Developers")]
+    mfg_name: &'static str,
+    /// Model number
+    #[characteristic(uuid = characteristic::MODEL_NUMBER_STRING, read, value = "EX-DIS")]
+    model_number: &'static str,
+}
+
+/// Run the BLE stack.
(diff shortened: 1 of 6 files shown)


alexmoon:
I think the CI server might need a kick...

PR #526, opened 2026-01-20
Generic-queue feature should be specified by binary crates only

ivmarkov:
Subject says it all, mostly.

Given that `bt-hci-linux` is a library, I don't think it has business with specifying the embassy-time generic queue size. This should be done by (the) binary crate using it.

A few other features which are a bit borderline for me as well:
- `std` on `embassy-time`. If I'm not mistaken, this enables the STD time driver. Shouldn't that be done by the binary crate as well? I mean, can't imagine some _other_ driver to be used with Linux, but still?
- `std` on `critical-section` - ditto

If you feel we should remove those too, let me know, I'll update the PR. I'm not 100% sure though hence why I did not touch those.

### DIFF
--- a/bt-hci-linux/Cargo.toml
+++ b/bt-hci-linux/Cargo.toml
@@ -15,7 +15,7 @@ tokio = { version = "1", features = ["io-util", "net", "rt", "macros", "signal",
 embedded-io = { version = "0.7", features = ["std"] }
 libc = "0.2"
 bt-hci = "0.7"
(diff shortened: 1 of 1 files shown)


ivmarkov:
If you are wondering, the context is that the binary crate in question (a large workspace, really) was using `generic-queue-64` which conflicted with the `generic-queue-8` of its `bt-hci-linux` dependency.

lulf:
I think keeping std is ok. If there is an alternative implementation for linux that people will use, let's change when the need arises.

PR #530, opened 2026-01-21
feat: Allow changing the value of a ReadOnlyData attribute

alexmoon:
This is useful if you have an attribute whose value can change between different large static values, but is read-only from the perspective of the client. For example, in my application we have an HID service whose report map characteristic can switch between a few different predefined values. These are very large values which are known at compile time, so it is convenient to keep them in flash and not use up RAM.

### DIFF
--- a/host/src/attribute.rs
+++ b/host/src/attribute.rs
@@ -363,6 +363,23 @@ impl<'d, M: RawMutex, const MAX: usize> AttributeTable<'d, M, MAX> {
         }
     }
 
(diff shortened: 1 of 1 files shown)


lulf (approved):

alexmoon:
I’m pretty sure this change can’t cause the tests to fail. Does something need to be reset?

lulf:
I forgot a plug in a USB cable in CI :)

PR #532, opened 2026-01-23
refactor: Use binary search for AttributeTable lookups

alexmoon:
This adds two new lookup methods to `AttributeTable`: `with_attribute(handle, fn)` and `iterate_from(start, fn)`. These simplify common patterns of attribute lookup and allow for some optimization. Since the attribute handles are guaranteed to be in sorted order, these functions use binary search instead of linear scan to find the handle. For large attribute tables this could be a significant speed improvement (especially on embedded where cache locality is less of a concern).


alexmoon:
After further thought, I think the approach in #533 is superior. I'll leave this up as a draft in case the 16-aligned service attribute handles is considered to be important enough to keep.

PR #540, opened 2026-02-05
feat: Add attribute permissions to AttributeTable

alexmoon:
This add first-class permissions to the attributes in the AttributeTable, separate from characteristic properties (as per Core Vol 3, Part F, 3.2.5). I've implemented a level system rather than a bitmask because there's no way in Trouble to have an authenticated connection that is not also encrypted.

### DIFF
--- a/examples/apps/src/ble_bas_peripheral_bonding.rs
+++ b/examples/apps/src/ble_bas_peripheral_bonding.rs
@@ -30,19 +30,19 @@ static DESC: [u8; 67] = [
 
 #[gatt_service(uuid = service::HUMAN_INTERFACE_DEVICE)]
 pub(crate) struct HidService {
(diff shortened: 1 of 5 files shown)


host/src/attribute.rs
```diff
@@ -39,6 +39,63 @@ pub enum CharacteristicProp {
...
+            }
+            PermissionLevel::AuthenticationRequired if level < SecurityLevel::EncryptedAuthenticated => {
+                Err(AttErrorCode::INSUFFICIENT_AUTHENTICATION)
+            }
+            _ => Ok(()),
+        }
+    }
+
+    pub(crate) fn can_write(&self, level: SecurityLevel) -> Result<(), AttErrorCode> {
+        match self.read {
```
HaoboGu: self.write?

alexmoon: Oops!

PR #543, opened 2026-02-08
fix: guard HostBufferSize behind controller-host-flow-control feature

Ic-arbon:
Hi! I'm a maintainer of [sifli-rs](https://github.com/OpenSiFli/sifli-rs), a Rust HAL for the SiFli SF32LB52x BLE SoC. While integrating trouble-host into our BLE stack, I ran into an issue where `ControlRunner::run()` exits immediately during initialization.

## Problem

The SF32LB52x's BLE controller does not support the `HCI_Host_Buffer_Size` command — it returns status `0x11` (Unsupported Feature). Since this command is currently called unconditionally with `?`, the error propagates and causes the control future to exit before any advertising commands are sent.

## Context

Looking at the git history, I noticed that in 833459b ("guard flow control behind feature"), both `HostBufferSize` and `SetControllerToHostFlowControl` were placed together inside `#[cfg(feature = "controller-host-flow-control")]`. After the refactor in #299, `HostBufferSize` ended up outside the cfg guard, and `SetControllerToHostFlowControl` was later commented out in #365.

Since `HostBufferSize` is a prerequisite for controller-to-host flow control (it tells the controller about the host's buffer capacity), it belongs behind the same feature flag.

## Changes

- Move `HostBufferSize` back inside `#[cfg(feature = "controller-host-flow-control")]`

## Test

Tested on SiFli SF32LB52x (ARM Cortex-M33):
- **Before**: `HostBufferSize` returns `0x11`, runner exits, no advertising
- **After**: Initialization succeeds, BLE advertising works correctly

Happy to discuss alternative approaches if unconditional `HostBufferSize` is preferred — for instance, we could ignore the error instead.

### DIFF
--- a/host/src/host.rs
+++ b/host/src/host.rs
@@ -1114,7 +1114,6 @@ impl<'d, C: Controller, P: PacketPool> ControlRunner<'d, C, P> {
         host.connections
             .set_link_credits(ret.total_num_le_acl_data_packets as usize);
 
-        #[cfg(feature = "controller-host-flow-control")]
         {
             const ACL_LEN: u16 = 255;
             const ACL_N: u16 = 1;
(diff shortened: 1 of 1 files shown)


lulf (changes requested):
Great to see more hw support! 

If I remember correctly, this command is still useful for the controller even if flow control is not explicitly enabled. I'd prefer to print a warning logging the error, and continue running.

PR #545, opened 2026-02-09
feat: Add auto-pts qualification tester firmware

alexmoon:
Adds a `tester` project which implements the Bluetooth Test Protocol of [`auto-pts`](https://github.com/auto-pts/auto-pts) to enable running PTS qualification tests against the trouble host stack.

This requires a number of small API changes in order to get access to the data and procedures needed by the tester. I've tried to make each commit a self-contained API change for ease of review.

The most significant change is to handling of connection parameter update requests which was needed in order to pass some mandatory GAP test cases.

### DIFF
--- a/host/src/gatt.rs
+++ b/host/src/gatt.rs
@@ -831,6 +831,18 @@ impl<'reference, C: Controller, P: PacketPool, const MAX_SERVICES: usize> GattCl
 
         let len = w.len();
         connection.send(Pdu::new(buf, len)).await;
+
+        // Await MTU exchange completion (BT Core Spec requires sequential ATT requests)
+        loop {
+            let pdu = connection.next_gatt_client().await;
+            match pdu.as_ref()[0] {
+                att::ATT_EXCHANGE_MTU_RSP | att::ATT_ERROR_RSP => break,
+                _ => {
+                    warn!("[gatt] unexpected PDU during MTU exchange, discarding");
+                }
+            }
+        }
+
         Ok(Self {
             known_services: RefCell::new(heapless::Vec::new()),
             stack,
--- a/host/src/host.rs
+++ b/host/src/host.rs
@@ -470,6 +470,8 @@ where
                 } else if let Ok(att::Att::Server(AttServer::Response(att::AttRsp::ExchangeMtu { mtu }))) = a {
                     info!("[host] remote agreed att MTU of {}", mtu);
                     self.connections.exchange_att_mtu(acl.handle(), mtu);
+                    #[cfg(feature = "gatt")]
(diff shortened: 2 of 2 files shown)


host/src/lib.rs
```diff
@@ -701,6 +702,11 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
     pub fn get_bond_information(&self) -> Vec<BondInformation, BI_COUNT> {
         self.host.connections.security_manager.get_bond_information()
     }
+
+    /// Get a connection by its peer address
+    pub fn get_connection(&'stack self, peer_address: Address) -> Option<Connection<'stack, P>> {
```
lulf: I think this should also be named get_connection_by_peer_address or get_peer_connection or something to make it very clear, because i don't think the variable naming is as visible as the function name.

alexmoon:
>     * How long does the test run with autopts take?

It's pretty slow, around 30 seconds per test case. My MVP test suite has 117 test cases. For comparison the Zephyr test suite has over 3800 test cases (but that includes a lot of stuff trouble will never support). I'd estimate a full test suite for a mature trouble to be around 500 test cases.

It can also be flaky. The PTS software doesn't seem to always recover from aborted tests properly. I've had to manually restart everything on the windows server and/or unplug and replug the dongle periodically to get it working again.

My suggestion would be to run the qualification tests on maybe a weekly basis instead of with every PR?

>     * Do I need to use two nrf52840 dk's or will nrf52832/nrf52833 work as well?

This is the list of dongles officially supported: https://support.bluetooth.com/hc/en-us/articles/360049018492-What-is-the-difference-between-the-Profile-Tuning-Suite-PTS-Dongles The Zephyr docs imply you can use a nrf52 dk, but I've never tried it. I think that the nrf54L15DK is the way to go since it supports Bluetooth 6.0. I've been using the Laird dongle (which is an nrf52840) since I had it from previous testing.

You can use whatever you want for the IUT. I've only implemented support for the nrf52840 so far but it should be easy to add the other chips.

At the moment I'm passing all the GAP tests in my test suite and most of the GATT server tests. GATT client is not yet working (will need some changes to trouble). L2CAP is not yet implemented in the tester.

alexmoon:
I've added more fixes based on qualification test failures. With this version I'm passing all GAP tests and nearly all GATT tests in my minimal test suite.

There is one change you need to make to auto-pts I forgot to mention. In autopts/ptsprojects/zephyr/gatt.py you need to change "TSPX_bearer_for_le" from "EATT" to "ATT" (line 120). At least until TrouBLE supports enhanced ATT bearers. Eventually we'll probably want our own ptsprojects setup instead of using zephyr's, but that's the only change I've needed to make so far.

PR #554, opened 2026-02-23
feat(security): Security manager fixes and additional features

alexmoon:
The main new feature is support for legacy pairing, which is required by the PTS test suite for some unrelated features (this is probably a bug in the test suite, but it was easier to just implement legacy pairing).

This also changes `GatClient::new` and `Characteristic::notify`/`Characteristic::indicate` to automatically attempt to enable encryption on connections with bonded peers. This enforces the Bluetooth requirement that encryption be established before attempting service requests with bonded peers.

I've also added a BondLost connection event when establishing encryption with a stored bond fails.

The other changes are various fixes to issues detected by the PTS test suite. There are still some more issues outstanding that need to be addressed, but the set of changes was getting pretty large so I'd like to get this reviewed first before tackling the remaining issues.

### DIFF
--- a/host/src/security_manager/pairing/central.rs
+++ b/host/src/security_manager/pairing/central.rs
@@ -424,12 +424,25 @@ impl Pairing {
             trace!("Handling {:?}, step {:?}", command.command, current_step);
             match (current_step, command.command) {
                 (Step::Idle, Command::SecurityRequest) => {
+                    // Parse the peer's AuthReq from the SecurityRequest payload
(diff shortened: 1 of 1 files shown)


alexmoon:
@lulf do you know if that test failure is real or spurious? I don't think the example tests use the security manager so I'd be surprised if this PR affected them.

lulf:
Im not close to the ci this week, but i think its not related

Issue #441, opened 2025-08-27
Macro-configurable GATT event dispatch


ericduvic:
### Purpose
To separate concerns and avoid cluttering the GattEvent loop, it would be nice to be able to specify handler functions for characteristics in the `characteristic` macro itself.

### Example
```
#[gatt_service(uuid = service::ENVIRONMENTAL_SENSING)]
pub struct TemperatureService {
    #[characteristic(uuid = characteristic::TEMPERATURE, read=get_temperature, notify)]
    pub temperature: i16,
    #[characteristic(uuid = CONTROL_TEMPERATURE_UUID, read, write=set_control_temperature)]
    pub control_temperature: i16,
}

impl TemperatureService {
    fn get_temperature() -> Result<i16, Error> {
        let result: i16 = 0;
        // Some code here to pull from an `embassy-sync` object, for example
        return result // the actual value to be returned from a `GattEvent::Read`
    }

    fn set_control_temperature(value: i16) -> Result<i16, Error> {
        // Some code here to update from an `embassy-sync` object, for example
        return value // Maybe deliver the actual value to be written to the object here if it needs to be mutated within this function
    }
}
```

ivmarkov:
I think this issue is another incarnation of "hey, I don't want to store the attribute data _inside_ the attribute table but in my own custom storage". One of the problems I had with the `rs-matter` BTP protocol which ended at least partially influencing the `GattEvent` code.

I think implementing such callbacks might make a lot of sense **if** we additionally:
- Make them mandatory, so that we can get rid of the variable-size-attribute-storage completely
- Make them async (because, why not, and if possible? might be an issue with `&dyn`)
- Will have to have - on the `AttributeTable` level - a generic pair of callbacks of the form "give me attribute XYZ value" and "update attribute XYZ value"
- Get rid of the [attribute storage inside the attribute table](https://github.com/embassy-rs/trouble/blob/366ee88a2aa19db11eb0707c71d797156abe23f5/host/src/attribute.rs#L81). 

Good side effects from this exercise:
- This would remove the `'values`  lifetime from `AttributeTable` and `AttributeServer`
[...]

PR #561, opened 2026-03-14
feat(gatt/server): GATT server fixes and features

alexmoon:
Added:
- `Stack::connections()` returns an iterator over currently connected connections.
- `AttributeServer::write_and_notify()` writes a characteristic in the `AttributeTable` and sends notifications/indications to all connected connections that have subscribed.
- `AttributeServer::read()` reads an attribute from the `AttributeTable` in the context of the provided connection. This means that client-specific attributes (like CCCD) will have the correct value and permission checks are performed. This is mostly relevant for the tester.

Fixed:
- Clear the CCCD table for a non-bonded connection when it disconnects.

With these changes all GATT tests in my PTS test suite are passing. In total: GAP, GATT, and SM are now all passing (with the exception of a couple of SM tests dealing with IRKs). L2CAP is the remaining area with significant failures.

### DIFF
--- a/tester/app/src/btp/mod.rs
+++ b/tester/app/src/btp/mod.rs
@@ -895,17 +895,30 @@ async fn handle_gatt<'a, 's, C: crate::Controller, P: PacketPool>(
             Ready(GattResponse::Attrs(attrs))
         }
         GetAttrValue(cmd) => {
+            let address = Address {
+                kind: cmd.addr_type,
+                addr: cmd.address,
+            };
             let mut buf = alloc::vec![0u8; 512];
-            match server.table().read(cmd.handle, 0, &mut buf) {
+            let result = match stack.get_connection_by_peer_address(address) {
+                Some(conn) => server.read(&conn, cmd.handle, 0, &mut buf),
+                None => server
+                    .table()
+                    .read(cmd.handle, 0, &mut buf)
+                    .map_err(|_| AttErrorCode::INVALID_HANDLE),
+            };
+            match result {
                 Ok(len) => {
                     buf.truncate(len);
-                    let data = buf.into_boxed_slice();
                     Ready(GattResponse::AttrValue(protocol::gatt::AttrValueResponse {
                         att_response: 0x00,
-                        value: data,
+                        value: buf.into_boxed_slice(),
                     }))
                 }
-                Err(_) => Error(BtpStatus::Fail),
+                Err(e) => Ready(GattResponse::AttrValue(protocol::gatt::AttrValueResponse {
+                    att_response: e.to_u8(),
+                    value: alloc::boxed::Box::default(),
+                })),
             }
         }
 


lulf (approved):
LGTM! Thanks again for all the work in fixing those autopts issues.

I've been attempting to refactor the GATT server a few times to support issue #441 , and one thing I keep running into is the CCCD tables. I'm wondering if they belong more as part of the connection manager, guarded behind the 'gatt' feature instead of part of the 'shared' gatt server. Thoughts on that?

The other thing is making the attribute table optional, dispatching into some async closures for read/writes to attributes. Is that something that still would be possible to do while complying with the GATT specification?

alexmoon:
> I've been attempting to refactor the GATT server a few times to support issue #441 , and one thing I keep running into is the CCCD tables. I'm wondering if they belong more as part of the connection manager, guarded behind the 'gatt' feature instead of part of the 'shared' gatt server. Thoughts on that?

I think moving them to the connection manager would be fine. In principle they're somewhat related to the bond information in that they need to be persisted between connections for bonded peers.

I also think we should consider generalizing the mechanism rather than making it CCCD specific. There is one other attribute in the core spec which is supposed to be handled in the same way (the Client Supported Features characteristic) and there may be others in specific GATT services. Perhaps have an AttributeData variant that is client-specific (e.g. `ClientSpecific { uuid: Uuid, permissions: AttPermissions }` in place of the `Cccd` variant)?

For #441 I'd probably suggest an AttributeData variant that holds a dyn Fn references for read and write.
 
> The other thing is making the attribute table optional, dispatching into some async closures for read/writes to attributes. Is that something that still would be possible to do while complying with the GATT specification?

I don't think async closures would be sufficient. You could potentially make things generic on an AttributeTable trait or use a dyn AttributeTable reference. The host needs at minimum to be able to iterate the table, look up attributes by handle, get UUIDs and permissions for attributes, and read and write attribute data. I'm not sure you'd gain much from that abstraction, though.

lulf:
> > I've been attempting to refactor the GATT server a few times to support issue #441 , and one thing I keep running into is the CCCD tables. I'm wondering if they belong more as part of the connection manager, guarded behind the 'gatt' feature instead of part of the 'shared' gatt server. Thoughts on that?
> 
> I think moving them to the connection manager would be fine. In principle they're somewhat related to the bond information in that they need to be persisted between connections for bonded peers.
> 
> I also think we should consider generalizing the mechanism rather than making it CCCD specific. There is one other attribute in the core spec which is supposed to be handled in the same way (the Client Supported Features characteristic) and there may be others in specific GATT services. Perhaps have an AttributeData variant that is client-specific (e.g. `ClientSpecific { uuid: Uuid, permissions: AttPermissions }` in place of the `Cccd` variant)?

Yes, I think that makes sense.

> 
> For #441 I'd probably suggest an AttributeData variant that holds a dyn Fn references for read and write.

I think this was there before I removed it again when introducing the gatt read/write event api. But I removed it because I didn't really like having multiple ways to do the same thing. I guess what I'm trying is to have one of the mechanisms build on top of the other.

> 
> > The other thing is making the attribute table optional, dispatching into some async closures for read/writes to attributes. Is that something that still would be possible to do while complying with the GATT specification?
> 
> I don't think async closures would be sufficient. You could potentially make things generic on an AttributeTable trait or use a dyn AttributeTable reference. The host needs at minimum to be able to iterate the table, look up attributes by handle, get UUIDs and permissions for attributes, and read and write attribute data. I'm not sure you'd gain much from that abstraction, though.

I was thinking about an approach where the attribute characteristic data and the value data was split. I.e. an AttributeMetadataTable that would look a bit like today, but somehow attaching closures for reading/writing the attributes that have values, similar to the dyn fn as you suggest. Then you could make the existing 'stored value' use the same mechanism under the hood. But yeah, I'm not sure what would be gained from that, possibly simplifying the Gatt WriteEvent/ReadEvent loop that's there today.

alexmoon:
> I was thinking about an approach where the attribute characteristic data and the value data was split. I.e. an AttributeMetadataTable that would look a bit like today, but somehow attaching closures for reading/writing the attributes that have values, similar to the dyn fn as you suggest. Then you could make the existing 'stored value' use the same mechanism under the hood. But yeah, I'm not sure what would be gained from that, possibly simplifying the Gatt WriteEvent/ReadEvent loop that's there today.

You could make `AttributeData` a trait with read and write methods. The `AttributeTable` would then become essentially `[(Uuid, AttPermissions, &'d dyn AttributeData<'d>); N]`. The existing variants would become `AttributeData` impls. But now you need a place to store the `AttributeData` values so I think it just ends up messier.

alexmoon:
One thing you could do to simplify things is remove all the special cased `AttributeData` variants. It would be up to the builder to populate the attribute table correctly. ~This would probably increase the size of the attribute table (permissions are no longer implicit for some variants and `SmallData` has to be increased from 8 to 20 bytes to handle characteristic declarations in-line).~ edit: It turns out the size stays the same thanks to padding bytes.

You also lose the service grouping data so you'd either have to iterate the table to find the last handle in a service group or store the grouping information elsewhere.

```rust
pub struct Attribute<'a> {
    pub(crate) uuid: Uuid,
    pub(crate) permissions: AttPermissions,
    pub(crate) data: AttributeData<'a>,
}

#[derive(Debug, PartialEq, Eq)]
pub(crate) enum AttributeData<'d> {
    ReadOnlyData {
        value: &'d [u8],
    },
    Data {
        variable_len: bool,
        len: u16,
        value: &'d mut [u8],
    },
    SmallData {
        variable_len: bool,
        capacity: u8,
        len: u8,
        value: [u8; 20],
    },
    ClientSpecific,
    Deferred {
        read: &'d mut dyn FnMut() -> Result<(), Error> + 'd,
        write: &'d mut dyn FnMut() -> Result<(), Error> + 'd,
    },
}
```

PR #562, opened 2026-03-18
fix(l2cap): Various L2CAP fixes

alexmoon:
Various L2CAP issues found via PTS testing. Mostly dealing with error conditions or edge cases.

With this change trouble can now pass my entire PTS test suite.

### DIFF
--- a/benchmarks/nrf-sdc/src/bin/l2cap_benchmark_central.rs
+++ b/benchmarks/nrf-sdc/src/bin/l2cap_benchmark_central.rs
@@ -4,8 +4,8 @@
 use defmt::{info, unwrap};
 use embassy_executor::Spawner;
 use embassy_futures::join::join;
-use embassy_nrf::peripherals::RNG;
 use embassy_nrf::mode::Async;
+use embassy_nrf::peripherals::RNG;
 use embassy_nrf::{bind_interrupts, rng};
 use embassy_time::{Duration, Instant, Timer};
 use nrf_sdc::mpsl::MultiprotocolServiceLayer;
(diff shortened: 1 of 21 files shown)


lulf (commented):
Thanks! Fantastic to get this tested with autopts. I will spend some time reviewing and testing with our workloads in the coming days.

alexmoon:
The integration test failure is because the test is using a PSM of 0x2349. However that is not a valid SPSM to use with a credit based connection (see Core Vol 3, Part A, 4.22). Do you want me to change it to 0x80 or some other value?

lulf:
Yes, change it to anything but 0x81 (which is used by example tests)

lulf:
However, I think that we should allow PSM outside that range. At least the old nrf-softdevice allows it.

alexmoon:
> However, I think that we should allow PSM outside that range. At least the old nrf-softdevice allows it.

I checked Zephyr's implementation and it also rejects PSMs outside of the valid range. Since it's not permissible under the spec to use those PSMs I think we should reject them. I'm currently using a 256-bit bitmask to track which PSMs are in use, that would need to change if we wanted to support illegal PSMs.

lulf:
> > However, I think that we should allow PSM outside that range. At least the old nrf-softdevice allows it.
> 
> I checked Zephyr's implementation and it also rejects PSMs outside of the valid range. Since it's not permissible under the spec to use those PSMs I think we should reject them. I'm currently using a 256-bit bitmask to track which PSMs are in use, that would need to change if we wanted to support illegal PSMs.

I agree it makes sense perhaps in the longer term, but there can be devices in the field relying on it. At least guard the accept/rejection/check behavior behind a feature flag (reject by default  or don't check it is ok, I'm not asking for another check implementation.), so that the previous behavior can be restored for those that already rely on it with their existing devices until they can be updated.

alexmoon:
I've added an allow-reserved-l2cap-psu feature to remove these checks. Note that with this feature enabled, a peer making an L2CAP connection request to a reserved PSU that is not being listened to will not get any response and the channel will remain permanently allocated in the channel manager. This matches the behavior of trouble prior to the other changes in this PR.

lulf (approved):
LGTM, thanks again!

PR #579, opened 2026-03-31
feat(security): Support for resolvable private addresses

alexmoon:
This add support for Resolvable Private Addresses, both for the local device address and for resolving peer addresses.

There are a few refactors that affect other parts of the crate:
- `trouble_host::new()` now returns a `StackBuilder`. You must call `build()` on the stack builder to get a `Stack`. This allows for clearer separation between build-time operations and run-time operations and allows those operations that can be done at either time (like `set_io_capabilities()`) to use an API appropriate to each phase. It also improves correctness by ensuring that only one `Runner` may be constructed for a `Stack`. `trouble_host::new()` takes a mutable reference to an `Option<Stack>` to allow the stack to be constructed in-place. `StackBuilder::build()` returns a tuple with a reference to the `Stack` and the `Runner`. `Central` and `Peripheral` now have accessor methods directly on `Stack`.
- `BondInformation` now stores a full `Address` instead of just a `BdAddr`. The `AddrKind` is needed for correctness in some pairing operations. As a result, the peer `AddrKind` is no longer stored separately in `ConnectionStorage`
- Similarly, `filter_accept_list` and `connect()` now take `Address`es instead of separate `AddrKind` and `BdAddr`s. Also `Connection::peer_address()` returns an `Address` and `peer_addr_kind()` is removed.
- Added `own_addr_kind` field to `AdvertisementParams` and an optional `address` to `AdvertisementSet` to allow controlling the address used when advertising.
- The "Central Address Resolution" characteristic is added to the GAP service when the `security` and `central` features are enabled. This indicates to peripherals we can resolve private addresses. I did not add the "Resolvable Private Address Only" characteristic, since that is dependent on the application. This will affect the required `AttributeTable` size for those applications enabling those features.

I also made a couple of changes to the `Controller` trait:
- I extracted the HCI commands required for the `security` feature into a separate `SecurityCmds` trait. This way those bounds are only required when the `security` feature is enabled.
- Similarly I added a `fmt::Format` trait which is a sub-trait of `defmt::Format` when `defmt` is enabled and a sub-trait of `Debug` otherwise. I added `fmt::Format` as a bound on the `Controller::Error` type so trouble can always log controller errors.

For private addresses, an application can use private addresses by providing an IRK to the `enable_privacy` method on `StackBuilder`. The address rotation interval can be controlled with the `set_rpa_timeout` method if desired, but defaults to 15 minutes. Resolvable private addresses of bonded peers will be automatically resolved by the controller using the resolving list. This should be completely transparent to the application.


host/src/lib.rs
```diff
@@ -651,30 +755,107 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
...
+    #[cfg(feature = "peripheral")]
+    pub fn peripheral(&'stack self) -> Peripheral<'stack, C, P> {
+        Peripheral::new(self)
+    }
+
+    /// Set the IO capabilities used by the security manager.
+    ///
+    /// Only relevant if the feature `security` is enabled.
+    #[cfg(feature = "security")]
+    pub fn set_io_capabilities(&self, io_capabilities: IoCapabilities) {
```
lulf, line 884: Does it make sense to be able to set these at run time, or should it be set on the builder instead? Should some action be taken in the security manager if this changed and it still has connections assuming other settings?

(Same question applies to the other settings below, I suppose)

lulf (commented):
Although I like adding a StackBuilder, I'm not sure about the 'in place' Option<Stack> in the builder API. Originally the HostResources contained the stuff that required a lot of space. Now it seems like the Stack is 'too big' (I assume, since you made this change). Should/could more of that reside in HostResources instead?

host/src/lib.rs
alexmoon: 99% of the time this should be set on the builder. However, the tester does need to be able to change I/O capabilities at runtime. Perhaps we should make it `#[doc(hidden)]`?

`set_secure_connections_only` had the same motivation, but I think with some recent changes to the tester I could remove the runtime version now. I'd need to run some tests to be sure.

I think changing the RPA timeout at runtime could be genuinely useful for some applications.

alexmoon:
> Although I like adding a StackBuilder, I'm not sure about the 'in place' Option in the builder API. Originally the HostResources contained the stuff that required a lot of space. Now it seems like the Stack is 'too big' (I assume, since you made this change). Should/could more of that reside in HostResources instead?

Looking at `-Z print-type-sizes`, the `BleHost` struct is currently 2088 bytes. Most of that comes from `ConnectionManager` which is 1808 bytes. And most of that comes from the `SecurityManager` which is 1568 bytes.

The other issue I was trying to solve with this design was to allow the `StackBuilder` to return the `Runner` alongside the `Stack` in order to enforce a single runner per stack. That required the `BleHost` to have a stable address that outlives the `StackBuilder`.

I went with `Option<Stack>` instead of `MaybeUninit<Stack>` to allow it to be safely dropped and re-built.

Perhaps we could put a `MaybeUninit<BleHost>` into `HostResources`? If we did that we'd probably need to move the connections/channels/etc into their respective managers to avoid self references.

lulf:
> > Although I like adding a StackBuilder, I'm not sure about the 'in place' Option in the builder API. Originally the HostResources contained the stuff that required a lot of space. Now it seems like the Stack is 'too big' (I assume, since you made this change). Should/could more of that reside in HostResources instead?
> 
> Looking at `-Z print-type-sizes`, the `BleHost` struct is currently 2088 bytes. Most of that comes from `ConnectionManager` which is 1808 bytes. And most of that comes from the `SecurityManager` which is 1568 bytes.
> 
> The other issue I was trying to solve with this design was to allow the `StackBuilder` to return the `Runner` alongside the `Stack` in order to enforce a single runner per stack. That required the `BleHost` to have a stable address that outlives the `StackBuilder`.
> 
> I went with `Option<Stack>` instead of `MaybeUninit<Stack>` to allow it to be safely dropped and re-built.
> 
> Perhaps we could put a `MaybeUninit<BleHost>` into `HostResources`? If we did that we'd probably need to move the connections/channels/etc into their respective managers to avoid self references.

This was the case before, but it prevented re-use of HostResources in a static, re-creating the BleHost (iirc it was rs-matter that needed this, and I'm fairly happy with that).

I think instead, maybe the security manager could be refactored to move things into HostResources similar to how connection manager etc. does that?

lulf:
> > > Although I like adding a StackBuilder, I'm not sure about the 'in place' Option in the builder API. Originally the HostResources contained the stuff that required a lot of space. Now it seems like the Stack is 'too big' (I assume, since you made this change). Should/could more of that reside in HostResources instead?
> > 
> > 
> > Looking at `-Z print-type-sizes`, the `BleHost` struct is currently 2088 bytes. Most of that comes from `ConnectionManager` which is 1808 bytes. And most of that comes from the `SecurityManager` which is 1568 bytes.
> > The other issue I was trying to solve with this design was to allow the `StackBuilder` to return the `Runner` alongside the `Stack` in order to enforce a single runner per stack. That required the `BleHost` to have a stable address that outlives the `StackBuilder`.
> > I went with `Option<Stack>` instead of `MaybeUninit<Stack>` to allow it to be safely dropped and re-built.
> > Perhaps we could put a `MaybeUninit<BleHost>` into `HostResources`? If we did that we'd probably need to move the connections/channels/etc into their respective managers to avoid self references.
> 
> This was the case before, but it prevented re-use of HostResources in a static, re-creating the BleHost (iirc it was rs-matter that needed this, and I'm fairly happy with that).
> 
> I think instead, maybe the security manager could be refactored to move things into HostResources similar to how connection manager etc. does that?

Thinking specifically about the `    bond: Vec<BondInformation, BOND_COUNT>` here

alexmoon:
The `SecurityManager` sizes break down like:
```
print-type-size type: `trouble_host::security_manager::Inner<10>`: 1328 bytes, alignment: 8 bytes
print-type-size     field `.state`: 600 bytes
print-type-size     field `.pairing_sm`: 304 bytes
print-type-size     field `.rng`: 312 bytes
print-type-size     field `.secret_key`: 32 bytes
print-type-size     field `.public_key`: 64 bytes
print-type-size     field `.finished_waker`: 8 bytes
print-type-size     field `.secure_connections_only`: 1 bytes
print-type-size     field `.io_capabilities`: 1 bytes
print-type-size     end padding: 6 bytes
```

The `state` field is mostly the bond information so that would help. But I think you'd probably want to move the pairing state machine as well, especially because I think that eventually trouble will need to support multiple pairing state machines. That will still leave the rng, which could potentially be passed in as a CryptoRng reference instead of our current seeding strategy?

Even with those changes, the `BleHost` struct will still be close to 900 bytes which I think is a bit heavy to be copying on the stack.

It also doesn't resolve the issue of how to enforce a single `Runner` instance per `Stack`.

alexmoon:
I should note these sizes come from `tester/nrf52`.

host/src/lib.rs
HaoboGu: just fyi: I just encountered a case which requires changing the I/O capabilities after `builder.build()` 2 days ago LOL: https://github.com/HaoboGu/rmk/pull/785#discussion_r3008278278 

In one word, I want to share the same stack building function across multiple binaries and the only difference between them is the I/O capabilities. So setting I/O capabilities after building is a "workaround".

lulf:
Hmm, good points, and thanks for checking the numbers. I suppose another alternative could be to have an additional 'level' of 'BleHostResources' holding the BleHost, ugh... Maybe the proposed Option<Stack> is better.

alexmoon:
One other thing to note about the `Option` approach is that you can put the `Option<Stack>` itself into a `StaticCell` or equivalent. And that does not prevent tearing down the stack because once the references go out of scope you can set the mutable reference back to `None` dropping the `Stack`.

I wonder if the better approach would be to just get rid of `HostResources` and recommend putting the `Option<Stack>` in a static instead?

edit: The main issue I see with this would be const generic pollution across the various types since we would no longer be using slices.

alexmoon:
I've put an alternate (and I think superior) `StackBuilder` API up in #581. If that is approved and merged, I'll rebase the private address support on top. I'll leave this PR as a draft until then.

alexmoon:
Rebased on #581. Ready for review.

lulf:
Apologies for the CI not working: I'm converting the machine to work more like bender (embassy CI)

lulf (approved):

PR #582, opened 2026-04-10
feat: setup bender

lulf:
Replace github actions with bender CI to speed up and allow more flexibility for configuration


lulf:
bender run

PR #584, opened 2026-04-13
fix(l2cap): Add L2capChannelListener

### LINKED ISSUE #563: L2capChannel::accept/listen race conditions
While working on #562 I noticed some potential race conditions with how `listen`/`accept` handle multiple connections to the same set of psms.

First of all, it’s possible for multiple connection requests to come in before the first has been accepted or rejected. There can be multiple connection request signals in the same PDU, or there could be multiple PDUs.

This can cause a couple of potential issues. First, if multiple requests are received before the `listen` future wakes, they will each allocate a channel and be in the `Connecting` state. This means if the user does not loop back and start listening again promptly after accepting or rejecting the first request the later requests will not receive a response and their channels will be left in a zombie state. 

Conversely, if the user wants to accept multiple connections to the same PSM, after `listen` returns the PSM will be marked unavailable until `listen` is called again which could cause connection requests to fail.

One idea would be to separate the listening state from the per-connection future. For example, we might have `L2capChannel::listen()` return a `L2capChannelListener` which would be responsible for registering and unregistering PSMs. It would have `next()` and `accept()` methods (returning `L2capPendingConnection` or `L2capChannel` respectively, equivalent to the current `listen` and `accept` methods). On drop it would unregister the PSMs and send rejects for any pending channels.

We might also consider whether listening should be scoped to a channel (as it is currently) or global to the whole stack. Most other hosts I've looked at register PSM servers globally to the stack, which means you don't have to spin up a listening task every time a connection connects. In the global model the user would have a single listening task, but then might need to spawn concurrent tasks to handle individual channels. We would presumably add a `Connection` accessor to `L2capChannel`.

alexmoon:
Fixes #563 by adding a `L2capChannelListener` that manages a `l2cap_listening` flag on the `Connection`. Also makes L2CAP PSM registration global so we don't have to track which SPSMs are registered for each connection. If finer-grained control is needed, users can register all SPSMs that might be supported then use `L2capPendingConnection::reject(LeCreditConnResultCode::SpsmNotSupported)` on a case-by-case basis.

I've also done a refactoring to rename all uses of "PSM" to "SPSM" to make it clear that we are using Simplified Protocol/Service Multiplexer values rather than the old BR/EDR Protocol/Service Multiplexer values.

Finally, I implemented a timeout for L2CAP channel connection requests per the spec.

### DIFF
--- a/host/src/channel_manager.rs
+++ b/host/src/channel_manager.rs
@@ -31,8 +31,8 @@ struct State {
     accept_waker: WakerRegistration,
     create_waker: WakerRegistration,
     disconnect_waker: WakerRegistration,
-    /// Bitmask tracking which PSMs (0x0001..=0x00FF) have active listeners.
-    registered_psms: [u32; 8],
+    /// Bitmask tracking which SPSMs (0x0001..=0x00FF) have active listeners.
+    registered_spsms: [u32; 8],
 }
 
 /// Channel manager for L2CAP channels used directly by clients.
@@ -76,24 +76,24 @@ impl<P, const QLEN: usize> PacketChannel<P, QLEN> {
 }
 
 impl State {
-    /// Register a PSM for listening.
-    fn register_psm(&mut self, psm: u16) {
-        if (1..=255).contains(&psm) {
-            let idx = psm as usize / 32;
-            let bit = psm as usize % 32;
-            self.registered_psms[idx] |= 1 << bit;
+    /// Register a SPSM for listening.
+    fn register_spsm(&mut self, spsm: u16) {
+        if (1..=255).contains(&spsm) {
+            let idx = spsm as usize / 32;
+            let bit = spsm as usize % 32;
+            self.registered_spsms[idx] |= 1 << bit;
         }
     }
 
-    /// Check if a PSM has an active listener.
-    fn is_psm_registered(&self, psm: u16) -> bool {
-        if !(1..=255).contains(&psm) {
+    /// Check if a SPSM has an active listener.
+    fn is_spsm_registered(&self, spsm: u16) -> bool {
+        if !(1..=255).contains(&spsm) {
             // When the allow-reserved-l2ca-psu feature is enable, treat all reserved PSUs as registered
             return cfg!(feature = "allow-reserved-l2cap-psu");
         }
-        let idx = psm as usize / 32;
-        let bit = psm as usize % 32;
-        self.registered_psms[idx] & (1 << bit) != 0
+        let idx = spsm as usize / 32;
+        let bit = spsm as usize % 32;
+        self.registered_spsms[idx] & (1 << bit) != 0
     }
 
     fn next_request_id(&mut self) -> u8 {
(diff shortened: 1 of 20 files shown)


host/src/l2cap.rs
```diff
@@ -261,44 +262,33 @@ impl<'d, P: PacketPool> L2capChannel<'d, P> {
...
         stack: &'d Stack<'_, T, P>,
-        connection: &Connection<'_, P>,
-        psm: &[u16],
-        config: &L2capChannelConfig,
-    ) -> Result<Self, BleHostError<T::Error>> {
-        let handle = connection.handle();
-        stack.host.channels.accept(handle, psm, config, stack.host).await
+        connection: &Connection<'d, P>,
+    ) -> Result<L2capChannelListener<'d, T, P>, Error> {
+        stack.host.connections.set_l2cap_listening(connection.handle(), true)?;
```
lulf: Since you have a reference to the connection, I think it would be better to use the index rather than the handle in this case for a more efficient lookup, so I would do this:

1. Add pub(crate) fn set_l2cap_listening  on Connection
2. Modify connection manager to take connection index instead
3. Call connection.set_l2cap_listening(true/false) here and in Drop

host/src/channel_manager.rs
```diff
@@ -472,9 +462,17 @@ impl<'d, P: PacketPool> ChannelManager<'d, P> {
...
+        let result = with_timeout(
+            L2CAP_RTX_TIMEOUT,
+            poll_fn(|cx| self.poll_created(conn, idx, ble, Some(cx))),
+        )
+        .await
+        .unwrap_or_else(|_| {
+            warn!("[l2cap] RTX timeout waiting for LE Credit Based Connection Response");
+            Err(BleHostError::BleHost(Error::Timeout))
+        })?;
         ondrop.defuse();
```
lulf, line 474: Previously I've handled this in the application code by wrapping the crate in with_timeout. I guess the forces the timeout to be within the host?

lulf (changes requested):
Overall looks good, thanks! I left one comment on the interaction with the connection manager when the connection index is available that I think should be fixed.

host/src/l2cap.rs
alexmoon: Yes, that's much better. Thanks!

host/src/channel_manager.rs
alexmoon: Yes, the spec requires the host to enforce a timeout on L2CAP signalling requests of between 1 and 60 seconds (Vol 3, Part A, 6.2). If a request times out, the host can either retry with an exponential timeout back-off or close the channel. I chose 30 seconds as a reasonable value which also matches the `ATT_TRANSACTION_TIMEOUT` value.

I believe all the relevant signalling requests we send are: `LeCreditConnReq`, `DisconnectionReq`, and `ConnParamUpdateReq`. `DisconnectionReq` doesn't need a timeout because we close the channel immediately. I couldn't figure out a good way to add a timeout to `ConnParamUpdateReq` because we don't wait for a response in `Connection::update_connection_params`. But adding a timeout to `LeCreditConnReq` seemed like an easy way to improve spec conformance.

lulf: Ok, sgtm.

lulf (approved):

PR #588, opened 2026-04-21
chore: add nrf52dk as test board

lulf:


lulf:
bender run

lulf:
bender run

PR #594, opened 2026-04-30
feat(gatt): Support arbitrary client-specific attributes

alexmoon:
This adds support for client-specific attributes beyond the CCCD descriptors.

I have simplified the `AttributeData` enum as originally outlined [here](https://github.com/embassy-rs/trouble/pull/561#issuecomment-4077859834). The custom variants for things like characteristic declarations and service definitions have been removed in favor of variants that just hold the raw data. The `Cccd` variant has been replaced with a general `ClientSpecific` variant for attributes that should have values that are unique per client and persisted across bonded connections to that client. The main down side is that we no longer have cached values for the last handle in a service or characteristic group. Those have been replaced with methods on `AttributeIterator` to scan the table for the next service or characteristic handle. Since the number of handles in any given service should be quite small this should have minimal performance impact.

Arbitrary length attributes are supported with a `ClientAttTable` struct that uses a "bag of bytes" to hold an index for the attribute locations within the data followed by the attribute data itself. This makes it easy to serialize and deserialize the table for storing across bonded connections.

I looked at moving the `ClientAttTable` into `ConnectionStorage`. That turned out not to work because the connections are completely independent of the `AttributeServer` until you create a `GattConnection`. That means there's nothing preventing you from having multiple `AttributeServer`s and using different ones for different connections (or event different ones at different times for the same connection) so having a single `ClientAttTable` tied to the connection doesn't make sense.

API changes:
- The `CCCD_MAX` generic parameter to `AttributeServer` is now `CLIENT_ATT_BYTES`. If your only client-specific attributes are CCCD descriptors then `CLIENT_ATT_BYTES` should be `2 + 6 * cccd_count`. The `#[server]` macro calculates this for you.
- `GattConnection::get` and `GattConnection::set` methods are added to get/set GATT values including client-specific value. These should generally be preferred over the generated server get/set functions because they handle client-specific values properly.
- `ServiceBuilder::add_characteristic_client` and `CharacteristicBuilder::add_descriptor_client` are added to allow creating client-specific attributes.
- `AttributeServer::get_cccd_table` and `AttributeServer::set_cccd_table` are renamed to `AttributeServer::get_client_att_table` and `AttributeServer::set_client_att_table`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ host-macros/Cargo.lock
 bt-hci-linux/Cargo.lock
 bt-hci-usb/Cargo.lock
 examples/apps/Cargo.lock
+examples/tests/bins
 tester/app/Cargo.lock
 .idea
 /tests/
--- a/docs/pages/gatt.adoc
+++ b/docs/pages/gatt.adoc
@@ -94,7 +94,7 @@ loop {
             match &event {
                 GattEvent::Read(event) => {
                     if event.handle() == level.handle {
-                        let value = server.get(&level);
+                        let value = conn.get(&level);
                         info!("Read battery level: {:?}", value);
                     }
                 }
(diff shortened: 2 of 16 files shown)


lulf (approved):
Nice. I think now that cccd table is a byte array, maybe it makes sense to expose this in config.rs to simplify the type signatures.

lulf:
bender run

PR #596, opened 2026-05-02
fix(host): use interval_max for max_connection_interval in LeRemoteConnectionParameterRequest handler

stonefoxes:
## Summary
- Fix copy-paste bug where `max_connection_interval` was incorrectly set from `event.interval_min` instead of `event.interval_max` in the `LeRemoteConnectionParameterRequest` handler (`host/src/host.rs:1348`).

This caused the host to always propose `min == max` connection intervals when handling remote connection parameter requests, ignoring the peer's actual requested range.


lulf (approved):
Thanks!

PR #598, opened 2026-05-04
add stm32wb example

xoviat:
confirmed on hardware


host/src/host.rs
```diff
@@ -1476,9 +1476,15 @@ impl<'d, C: Controller, P: PacketPool> ControlRunner<'d, C, P> {
...
+        if let Err(e) = SetEventMaskPage2::new(EventMaskPage2::new().enable_encryption_change_v2(true))
             .exec(&host.controller)
-            .await?;
+            .await
+        {
+            match e {
+                cmd::Error::Hci(bt_hci::param::Error::UNKNOWN_CMD) => debug!("set event mask page 2 is not supported"),
+                e => Err(e)?,
+            }
+        }
```
xoviat, line 1488: @lulf Don't merge just yet, but I wanted your thoughts about this

lulf: I would perhaps prefer it was a warning, but other than that I can't think of a way other to do this other than feature flags.

lulf (commented):
Looks good overall, great to have an stm32 example. Thank you!

HaoboGu (commented):
`ble_blas_peripheral.rs` -> `ble_bas_peripheral.rs`?

PR #599, opened 2026-05-07
feat(gatt): Make ClientAttTable size a config value instead of generic

alexmoon:
This make the ClientAttTable size a config.rs value instead of a const generic as suggested [here](https://github.com/embassy-rs/trouble/pull/594#pullrequestreview-4235765564). This reduces the number of generic parameters to `AttributeServer` by one which is nice.

In order to allow using persisted `ClientAttTable`s of a different size I've added a `ClientAttTableView` which wraps an arbitrarily sized byte slice.

I also changed the format of the table slightly to include the total `values_len` in the header. This simplifies the validation of tables and the calculation of the length of the last value in the table.


lulf (approved):
Thanks!

PR #601, opened 2026-05-10
Add `Encrypted` connection event

HaoboGu:


HaoboGu:
bender run

alexmoon:
I think this needs some careful design. The user application needs to be able to distinguish the following scenarios:
- Encryption is established, but the bonding procedure is still in progress
- Encryption is established, pairing is complete with no bond established
- Encryption is established to a bonded peer

`conn.is_bonded_peer()` may be able to distinguish the third, and otherwise you could wait for the `PairingComplete` event. However, that feels a bit fragile: if bonding to a new peer completes before you handle the  `Encrypted` event you would get a false positive.

HaoboGu:
> I think this needs some careful design. The user application needs to be able to distinguish the following scenarios:
> 
> * Encryption is established, but the bonding procedure is still in progress
> * Encryption is established, pairing is complete with no bond established
> * Encryption is established to a bonded peer

Maybe we should add a `BondResumed` event, make `PairingComplete` emitted only for fresh connections?

And for the first two scenarios, I think the order of emitted event should be fixed - `Encrypted` should be emitted first and then `PairingComplete`/`BondResumed`/`PairingFailed`.

lulf:
bender run

PR #602, opened 2026-05-10
Remove redundant `disconnected` call

HaoboGu:
When LTK is missing, `(self.disconnected(handle, Status::AUTHENTICATION_FAILURE)` is redundant because the HCI `Disconnect` command has been sent.

Local connection state is already cleared by the `DisconnectionComplete` handler once the controller acks the `Disconnect` HCI command, `disconnected` is called in the handler, it shouldn't be called here. Otherwise a second `disconnected` will be called which reports `"[link][disconnect] connection handle ... not found"`

### DIFF
--- a/host/src/connection_manager.rs
+++ b/host/src/connection_manager.rs
@@ -892,11 +892,9 @@ impl<'d, P: PacketPool> ConnectionManager<'d, P> {
     ) -> Result<(), crate::BleHostError<C::Error>>
     where
         C: crate::ControllerCmdSync<bt_hci::cmd::le::LeLongTermKeyRequestReply>
-            + crate::ControllerCmdAsync<bt_hci::cmd::le::LeEnableEncryption>
-            + crate::ControllerCmdSync<bt_hci::cmd::link_control::Disconnect>,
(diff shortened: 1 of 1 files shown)


alexmoon:
The connection's state should probably be updated to `Disconnecting(AUTHENTICATION_FAILURE)` after sending the disconnect command.

HaoboGu:
> The connection's state should probably be updated to `Disconnecting(AUTHENTICATION_FAILURE)` after sending the disconnect command.

Should `self.request_handle_disconnect(handle, DisconnectReason::AuthenticationFailure);` be used here, instead of using HCI command?

lulf:
> > The connection's state should probably be updated to `Disconnecting(AUTHENTICATION_FAILURE)` after sending the disconnect command.
> 
> Should `self.request_handle_disconnect(handle, DisconnectReason::AuthenticationFailure);` be used here, instead of using HCI command?

Yes, I think disconnects should go through that path.

HaoboGu:
Changed to use `request_handle_disconnect`

lulf:
bender run

Issue #603, opened 2026-05-10
SecurityManager crashes when connecting to multiple bonded peers


happenslol:
Ran into this while hacking on my own split keyboard firmware and attempting to pair the central half with both the peripheral and the host/computer. I'm not super familiar with BLE in general, so please excuse any inaccuracies.

From what I can tell, the issue is that `pairing_sm` in the `SecurityManager` (or rather `Inner`) is only ever cleared on disconnect, and not every interaction with it checks whether the peer address matches. So it's possible for pairing to fail because of a stale `pairing_sm` from another connection.

I'm working with 2 xiao nrf52840 controllers, if it helps I can provide a minimal example for those. Here's a general description of what I'm doing:

1. Have 2 bonded peers, a) one central and b) one peripheral
2. Initiate connection to peripheral a with stored bond and call `request_security()`, encryption succeeds
3. Peer for b initiates connection to us, requests security, encryption also succeeds
4. Connection is dropped, ble task crashes

After 2., `pairing_sm` is `Some(central_fsm { peer: a, step: Success })`. When encryption then succeeds in 3., we never check the address in `handle_encryption_success`, so `handle_event(pairing::Event::LinkEncryptedResult(true))` is called while the step is success, which results in `InvalidState`.

I'm not sure what the right fix is. We could just check the peer address of the state machine in `handle_encryption_success/failure`, and not do anything if it doesn't match. Re-encryption doesn't need the pairing state machine at all, from what I can tell. Storing a state machine per peer address seems formally correct to me, but I don't even know if multiple pairings can happen at the same time?

Happy to create a PR for this if you give me some pointers!

PR #606, opened 2026-05-13
fix(security): update peer identity when bond is finalized

rcloran:
When reconnecting a peer that completed bonding earlier in the same instance of the host stack, the stack would crash with `InvalidState`. The reason for this was that pairing state was not properly cleared during disconnect, and the reason for that was that the disconnect procedure compared the identity address in the pairing state machine to the RPA initially received at connection.

This change updates ConnectionManager's storage with the identity address of the peer once bonding is complete so that disconnect properly clears pairing state, allowing reconnect to succeed.

### DIFF
--- a/host/src/connection_manager.rs
+++ b/host/src/connection_manager.rs
@@ -959,10 +959,16 @@ impl<'d, P: PacketPool> ConnectionManager<'d, P> {
             }
             crate::security_manager::SecurityEventData::TimerChange => (),
             #[cfg(feature = "security")]
-            crate::security_manager::SecurityEventData::BondAdded(identity) => {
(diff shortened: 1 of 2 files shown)


rcloran:
I'm not sure if BondAdded is the right event to store this at. That is, I haven't tested or really thought through what happens for un-bonded pairing. It may be that we need to add a new event, but maybe it's not needed.

I think this is a more thorough attempt at a fix for the same problem that the first part of #597 tries to fix. 

#603 looks related, but different.

rcloran:
I suspect there's something similar necessary to update gatt servers on PairingComplete, but I can't figure out how to test that. Just a `self.server.update_identity(bond.identity)` looks like it should do the right thing.

HaoboGu (approved):

Issue #232, opened 2025-01-11
MTU handshake results in default size


pperanich:
When running the ble_bas_peripheral example on the Microbit, the [agreed upon MTU](https://github.com/embassy-rs/trouble/blob/6b464147efe6484da1d7ca2ea78247adaa393b13/host/src/host.rs#L355) results in the BLE default of 23 bytes, instead of the requested 247 that is specified `L2CAP_MTU - 4`. When running a similar application with `nrf-softdevice` (with the same devices for gatt server/client), the MTU handshake results in the upper limit being selected, i.e. whatever `ATT_MTU` sets.

pperanich:
Looking through the code base, it appears that neither the `Connection` nor `ConnectionManager`'s `set_att_mtu` method is called anywhere. When I change the visibility of this method to `pub` and add the following snippet to the ble_bas_peripheral example, the agreed MTU is larger, though still not the requested `ATT_MTU`. I.e it ends up being 185, instead of 247:

```rust
    info!("[adv] advertising");
    let conn = advertiser.accept().await?;
    conn.set_att_mtu(ATT_MTU as u16);  // <-- Add here.
    info!("[adv] connection established");
    Ok(conn)

```

If I add the following log to `exchange_att_mtu`:

```rust
    pub(crate) fn exchange_att_mtu(&self, conn: ConnHandle, mtu: u16) -> u16 {
        let mut state = self.state.borrow_mut();
        for storage in state.connections.iter_mut() {
            match storage.state {
                ConnectionState::Connected if storage.handle.unwrap() == conn => {
                    info!("storage.att_mtu = {:?}, mtu = {:?}", storage.att_mtu, mtu);  // <-- Added log
                    storage.att_mtu = storage.att_mtu.min(mtu);
                    return storage.att_mtu;
                }
                _ => {}
            }
        }
        mtu
    }
```

I get:
```
0.007110 INFO  [adv] advertising
└─ trouble_example_apps::ble_bas_peripheral::advertise::{async_fn#0} @ /Users/peranpl1/Documents/repos/oss/trouble/examples/apps/src/fmt.rs:143
11.909423 TRACE [host] connection with handle ConnHandle(0) established to BdAddr([b2, cf, 56, d5, 2c, 59])
└─ trouble_host::host::{impl#1}::handle_connection @ /Users/peranpl1/Documents/repos/oss/trouble/host/src/fmt.rs:117
[...]

PR #608, opened 2026-05-14
fix(att): apply MTU exchange while connection is still Connecting

andresv:
# Problem
`exchange_att_mtu` only matched `ConnectionState::Connected`. When the central's `ATT_EXCHANGE_MTU_REQ` arrived while the storage was still in `Connecting`, the match fell through. The function returned the peer's MTU unchanged so the caller's [host] agreed att MTU and log claimed success, but `storage.att_mtu` was silently left at `23`. For the lifetime of that connection, `gatt::assemble` then truncated every notification > `20` bytes.

Whether the request lands in `Connecting` or `Connected` seems to be scheduling race between the runner task and the application's `accept()` future after `LE Connection Complete`. Sometimes it worked, sometimes not.

# Fix
Add both: `Connecting | Connected`, matching the rest of trouble's ACL ingress (`connection_by_handle_mut`).

Might be fix also for [#232](https://github.com/bislydev/aurora/issues/232).

# Test
Silicon Labs EFR32MG26 + bluez 5.66 on a Raspberry Pi. Runs a chunked write/notify protocol where each END notification carries 35 bytes (3-byte status + 32-byte SHA-256).

## Before the patch:

`defmt log: [host] agreed att MTU of 517` (the fall-through value)
on-air via btmon: `Handle Value Notification len 22 / Data[20]: 0301..` - last 15 bytes of SHA-256 lost
host-side: SHA-256 mismatch, protocol fails

## After the patch:

`defmt log: [host] agreed att MTU of 247` (= default_att_mtu.min(peer))
on-air: full 35-byte notification
host-side: all blob hashes match

### DIFF
--- a/host/src/connection_manager.rs
+++ b/host/src/connection_manager.rs
@@ -659,9 +659,13 @@ impl<'d, P: PacketPool> ConnectionManager<'d, P> {
         debug!("exchange_att_mtu: {}, current default: {}", mtu, state.default_att_mtu);
         let default_att_mtu = state.default_att_mtu;
         core::mem::drop(state);
+        // Both states are live link-layer connections; the transition
+        // to `Connected` is just the application polling `accept()` and
+        // is unrelated to ATT activity. `handle_acl` already accepts
+        // both via `connection_by_handle_mut`.
         for storage in self.connections.borrow_mut().iter_mut() {
             match storage.state {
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks!

PR #610, opened 2026-05-17
fix: configure docs.rs build target and features

### LINKED ISSUE #604: Security API (BondInformation, add_bond_information, etc) not visible on embassy.dev or docs.rs
Ran into this implementing bonding, many of the APIs gated behind the `security` feature couldn't be found on embassy.dev or docs.rs.

I suspect it has to do with this part in the Cargo.toml, which doesn't include the `security` feature:
```toml
[package.metadata.embassy_docs]
src_base = "https://github.com/embassy-rs/trouble-host/blob/v$VERSION/src/"
src_base_git = "https://github.com/embassy-rs/trouble-host/blob/$COMMIT/src/"
features = ["defmt", "peripheral", "central", "gatt", "scan", "derive"]
target = "thumbv7em-none-eabi"
```

It would be great if the `security` feature (and other non-default features) could be added to this list, and if a `[package.metadata.docs.rs]` section could be added so docs.rs picks them up too!

GuyOnWifi:
Fixes #604

Adds the correct features and target so that docs.rs can pick it up. Previously, docs.rs was only building with default features leading to many APIs gated behind non-default features hidden. Also sets the target to match the embassy docs (thumbv7em-none-eabi).

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -17,6 +17,10 @@ src_base_git = "https://github.com/embassy-rs/trouble-host/blob/$COMMIT/src/"
 features = ["defmt", "peripheral", "central", "gatt", "scan", "derive", "security"]
 target = "thumbv7em-none-eabi"
 
+[package.metadata.docs.rs]
+features = ["defmt", "peripheral", "central", "gatt", "scan", "derive", "security"]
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks!

PR #613, opened 2026-05-22
Allow outbound att exchange replies when in connecting state

### LINKED ISSUE #611: ACL data dropped causing silent PDU loss on fast-connecting peers
## Bug description

When a BLE peer connects and immediately sends an ATT packet e.g., an MTU
exchange request, before the application task has had a chance to call
`accept()`, the packet is silently dropped. The result is a connection that
appears established on both sides but is functionally dead, the peer never
receives a reply to its first ATT request and the GATT setup sequence stalls,
at least from my experience.

## To Reproduce

Observed on an ESP32-C6 running an Embassy-based BLE peripheral. The device
advertises and accepts a single peripheral connection. The connecting peer, an
Android mobile app, sends an ATT MTU exchange request within ~1 ms of the HCI
`LeConnectionComplete` event being processed with the following logs.

```
[host] encountered error processing ACL data for ConnHandle(0): NotFound
```

## Environment

`trouble-host 0.6.0` but also reproduced against current `main`.

## (Possible) Root cause

This *might* be related to the fact that `post_gatt()` and `post_gatt_client()`
require the `Connected` state without buffering any ACL packets received during
the `Connecting` state and/or accepting `Connecting` state as well.

lulf:
Fixes two issues:

* Allowing tx runner to send outbound att exchange replies even if connection is not accept'ed.
* Enqueue inbound ATT requests if connection is not accepted.

I think this behavior is more desireable. If a connection times out, it is anyway cleaned up. If an application never accepts connections, thats a bug in the application and it's only normal to run out of resources.

Should fix https://github.com/embassy-rs/trouble/issues/611


renkenono:
@lulf the changes look good to me, thanks!

lulf:
bender run

lulf:
bender run

PR #615, opened 2026-05-29
fix: FIND_INFORMATION format byte wrong for 128-bit UUIDs

kokroo:
small ATT bug. the FIND_INFORMATION_RSP Format byte should be 0x01 for 16-bit UUIDs and 0x02 for 128-bit (Core spec Vol 3 Part F, 3.4.3.2), but `handle_find_information` writes `Uuid::get_type()` straight into it. get_type() returns 0x03 for 128-bit (it's a uuid-kind tag, not the ATT format), so descriptor discovery over a 128-bit characteristic emits Format=0x03, which is reserved.

bluez/android shrug it off, WinRT doesn't — on windows `discoverServices` just fails with a generic "Unknown error" (hit it with both flutter_blue_plus and bleak against an nrf52 peripheral).

repro is basically any service with a 128-bit char that has a descriptor (e.g. a CCCD from `notify`):

```rust
#[gatt_service(uuid = "12345678-1234-5678-1234-56789abcdef0")]
struct Svc {
    #[characteristic(uuid = "12345678-1234-5678-1234-56789abcdef1", read, notify)]
    x: [u8; 4],
}
```

FIND_INFORMATION over that range returns Format=0x03 today, should be 0x02. one-liner, just maps the kind tag to the real format code.

### DIFF
--- a/host/src/attribute_server.rs
+++ b/host/src/attribute_server.rs
@@ -561,7 +561,13 @@ impl<'values, M: RawMutex, P: PacketPool, const ATT_MAX: usize, const CONN_MAX: 
             }
             Ok::<(), codec::Error>(())
         })?;
-        header.write(t)?;
+        // FIND_INFORMATION_RSP "Format" byte (Core spec, Vol 3 Part F, 3.4.3.2):
+        // 0x01 = handle(s) + 16-bit UUID(s), 0x02 = handle(s) + 128-bit UUID(s).
+        // `Uuid::get_type()` returns a uuid-kind tag (16-bit => 0x01, 32-bit =>
+        // 0x02, 128-bit => 0x03), which is NOT the ATT format code, so writing it
+        // straight emits Format=0x03 (reserved) whenever the response carries
+        // 128-bit UUIDs. Map it to the spec value.
+        header.write(if t == 0x01 { 0x01u8 } else { 0x02u8 })?;
 
         if body.len() > 2 {
             Ok(header.len() + body.len())
--- a/host/src/attribute_server.rs
+++ b/host/src/attribute_server.rs
@@ -547,27 +547,29 @@ impl<'values, M: RawMutex, P: PacketPool, const ATT_MAX: usize, const CONN_MAX: 
         self.att_table.iterate_from(start, |it| {
             for (handle, att) in it {
                 if handle <= end {
+                    // FIND_INFORMATION_RSP "Format" byte (Core spec, Vol 3 Part F, 3.4.3.2):
+                    // 0x01 = handle(s) + 16-bit UUID(s), 0x02 = handle(s) + 128-bit UUID(s).
+                    // `get_att_type()` returns the real format code (32-bit reports 0x02 and is
+                    // widened to 128-bit by `att_bytes()`), so the header byte, this same-type
+                    // batching check, and the body written below all agree.
+                    let format = att.uuid.get_att_type();
                     if t == 0 {
-                        t = att.uuid.get_type();
-                    } else if t != att.uuid.get_type() {
+                        t = format;
+                    } else if t != format {
                         break;
                     }
-                    if body.available() < 2 + att.uuid.as_raw().len() {
+                    let (uuid, len) = att.uuid.att_bytes();
+                    let uuid = &uuid[..len];
+                    if body.available() < 2 + uuid.len() {
                         break;
                     }
                     body.write(handle)?;
-                    body.append(att.uuid.as_raw())?;
+                    body.append(uuid)?;
                 }
             }
             Ok::<(), codec::Error>(())
         })?;
-        // FIND_INFORMATION_RSP "Format" byte (Core spec, Vol 3 Part F, 3.4.3.2):
-        // 0x01 = handle(s) + 16-bit UUID(s), 0x02 = handle(s) + 128-bit UUID(s).
-        // `Uuid::get_type()` returns a uuid-kind tag (16-bit => 0x01, 32-bit =>
-        // 0x02, 128-bit => 0x03), which is NOT the ATT format code, so writing it
-        // straight emits Format=0x03 (reserved) whenever the response carries
-        // 128-bit UUIDs. Map it to the spec value.
-        header.write(if t == 0x01 { 0x01u8 } else { 0x02u8 })?;
+        header.write(t)?;
 
         if body.len() > 2 {
             Ok(header.len() + body.len())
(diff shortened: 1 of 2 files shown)


alexmoon:
This won't work properly if 32-bit UUIDs are present because they'll be considered a different type when considering whether to continue the loop but the same type in the response pdu.

`Uuid::get_type` is only used here, so I would be inclined to rename it to something like `get_att_type` and return the correct value there. Then when serializing the Uuid you need to be sure that 32-bit UUIDs are serialized as 16-byte (128-bit) values over the air.

davertay-j:
Confirming this from a different direction, and correcting one point in the description: BlueZ does **not** shrug off the invalid Format byte.

`btmon` on a BlueZ central, against a peripheral whose service uses a 128-bit UUID:

```
ATT: Find Information Response (0x05) len 19
        Format: unknown (0x03)
        0f 00 18 d2 03 7f 78 9d b6 90 86 4b 37 46 4f 33 7b ad
```

BlueZ does not recognise Format 0x03 and falls back to parsing the payload as 16-bit handle/UUID pairs, so the descriptor walk yields handles that are really byte pairs of the service's own 128-bit UUID:

```
handle = 0x000f, uuid = 00000000-0000-0000-0000-000000000000
handle = 0xd218, uuid = 00000000-...
handle = 0x7f03, uuid = 00000000-...
handle = 0x4b86, uuid = 00000000-...
```

`0xd218`, `0x7f03` and `0x4b86` are byte pairs of the real service UUID, `ad7b334f-4637-4b86-90b6-9d787f03d218`.

That makes the user-visible impact broader than a WinRT rejection: on BlueZ a client cannot discover descriptors at all on a service with 128-bit UUIDs, including the CCCD, so it cannot subscribe to notifications by discovery. Services using only 16-bit UUIDs are unaffected, which is presumably why this can go unnoticed.

Observed on trouble-host 0.7.0 against a real controller rather than the mock — a Dialog DA1469x-class controller over a `bt-hci` transport — and the relevant lines are unchanged on `main` at c21b1239.

On the outstanding review point: it is correct, and it does apply to the mapping in this PR as written. `t` comes from `get_type()`, so a run of 32-bit UUIDs yields `t == 0x02`, takes the `else` branch and emits Format 0x02, while the body writes `att.uuid.as_raw()` — 4 bytes for `Uuid::Uuid32`. A client would then try to read 18-byte entries out of a payload with 6 bytes per entry. The suggested shape — return the real ATT format code from the accessor, and serialise 32-bit UUIDs as 128-bit on air — resolves both halves, the Format byte and the loop's same-type comparison, since 32-bit and 128-bit entries then agree in both places.

One data point that may help scope that: on `main` at c21b1239, `get_type()` is defined only at `host/src/types/uuid.rs:94` and called only at `host/src/attribute_server.rs:551-552`, i.e. nowhere outside `handle_find_information`. So redefining it to return the ATT format code is contained. The thing to keep in mind is just that those two adjacent calls are the format byte and the same-type comparison, which is precisely why the 32-bit serialisation has to change along with it.

davertay-j:
We implemented @alexmoon's suggested shape and validated it on hardware, so offering it here in case it is useful — the design is entirely from the review comment above, we only wrote and tested it.

Two parts, and the second one is the load-bearing half:

1. The accessor returns the ATT Format code itself, renamed `get_att_type()`, so the header byte and the `t` same-type comparison come from one place and `header.write(t)` needs no change.
2. A companion `att_bytes()` expands a 32-bit UUID against the Bluetooth base UUID, so the body matches the `0x02` the header promises.

Part 2 is what makes part 1 safe, and it is easy to underestimate. We first tried it without the expansion, using a helper that returned `Option<u8>` and `None` for `Uuid32`, on the reasoning that §3.4.3.2 has no 32-bit form so such an attribute simply cannot be reported. That is well-formed on air but it silently truncates discovery: the `None` breaks the batch, and if the range *begins* at a 32-bit attribute the response ends up empty and becomes `ATTRIBUTE_NOT_FOUND`, hiding the attribute from the client altogether. The mapping in this PR as written has the mirror-image problem — Format `0x02` with 4-byte entries from `as_raw()`. Expanding to the base UUID is the only variant we found where the Format byte, the batching comparison and the body all agree.

```diff
--- a/host/src/types/uuid.rs
+++ b/host/src/types/uuid.rs
-    /// Get the UUID type.
-    pub fn get_type(&self) -> u8 {
+    /// Get the ATT Find Information Response format code for the UUID.
+    ///
+    /// Core Spec Vol 3 Part F, 3.4.3.2 defines only 0x01 (handle + 16-bit UUID pairs) and 0x02
+    /// (handle + 128-bit UUID pairs). A 32-bit UUID has no form of its own, so it reports 0x02 and
+    /// is written as its 128-bit equivalent; see [`Uuid::att_bytes`].
+    pub fn get_att_type(&self) -> u8 {
         match self {
             Uuid::Uuid16(_) => 0x01,
-            Uuid::Uuid32(_) => 0x02,
-            Uuid::Uuid128(_) => 0x03,
+            Uuid::Uuid32(_) | Uuid::Uuid128(_) => 0x02,
         }
     }
+
+    /// The UUID as an ATT Find Information Response carries it, and its length: two bytes under
+    /// format 0x01, sixteen under 0x02.
+    ///
+    /// A 32-bit UUID is expanded against the Bluetooth base UUID, which is the only way the 0x02
+    /// [`get_att_type`](Self::get_att_type) reports for it can be honoured on air.
+    pub(crate) fn att_bytes(&self) -> ([u8; 16], usize) {
+        match self {
+            Uuid::Uuid16(uuid) => {
+                let mut bytes = [0; 16];
+                bytes[..2].copy_from_slice(uuid);
+                (bytes, 2)
+            }
+            Uuid::Uuid32(uuid) => (
+                BluetoothUuid128::base()
+                    .set_initial_group(u32::from_le_bytes(*uuid))
+                    .to_le_bytes(),
+                16,
+            ),
+            Uuid::Uuid128(uuid) => (*uuid, 16),
+        }
+    }
```

```diff
--- a/host/src/attribute_server.rs
+++ b/host/src/attribute_server.rs
@@ handle_find_information
                 if handle <= end {
-                    if t == 0 {
-                        t = att.uuid.get_type();
-                    } else if t != att.uuid.get_type() {
+                    let format = att.uuid.get_att_type();
+                    if t == 0 {
+                        t = format;
+                    } else if t != format {
                         break;
                     }
-                    if body.available() < 2 + att.uuid.as_raw().len() {
+                    let (uuid, len) = att.uuid.att_bytes();
+                    let uuid = &uuid[..len];
+                    if body.available() < 2 + uuid.len() {
                         break;
                     }
                     body.write(handle)?;
-                    body.append(att.uuid.as_raw())?;
+                    body.append(uuid)?;
                 }
```

`BluetoothUuid128` is already imported in `types/uuid.rs`, so no new dependency, and `base().set_initial_group(..)` is the derivation from Vol 3 Part B. Two things worth flagging rather than hiding: `get_type()` is `pub`, so renaming it is technically a breaking change even though nothing in-tree calls it — adding `get_att_type()` alongside it would avoid that if you would rather. And `att_bytes()` is deliberately scoped to this response; other PDUs keep using `as_raw()`, since the Format-byte contract is what forces the widening here.

Hardware corroboration, which is the part we can actually add: this runs on a Dialog DA1469x with a `trouble-host`-based host, and with it service discovery from a BlueZ central completes correctly, including the descriptor walk that previously returned the garbage handles in my earlier comment. To be precise about the limits of that, though — we exercised **128-bit** UUIDs on real hardware. Our own abstraction layer above `trouble-host` has only 16-bit and 128-bit UUID variants, so a `Uuid32` cannot reach the attribute table through it and we have **not** hardware-tested the 32-bit path. That half rests on reading the spec and the `bt-hci` derivation, not on a capture.

Happy to open a PR against this branch or a fresh one if that is easier to review than a comment — just say which you'd prefer and I'll do that rather than leave it here.

kokroo:
I’ve pushed a follow-up that takes this from the header-byte-only fix to the full one @alexmoon described.
What changed:

1.  get_type() → get_att_type(), now returning the real ATT format code (16-bit → 0x01, 32-bit and 128-bit → 0x02) from a single place, so the header byte and the same-type batching check can’t disagree.
2. New att_bytes() serialises the UUID as the response actually carries it — 2 bytes under format 0x01, 16 under 0x02 — expanding a 32-bit UUID against the Bluetooth base UUID. That fixes the second half @alexmoon flagged: previously a Uuid32 emitted a 4-byte body under a 0x02 header. Now the Format byte, the batching comparison, and the body all agree.
3. Added a lib test covering both the 128-bit (Format 0x02, UUID echoed back verbatim) and 32-bit (widened to the base UUID) paths.

get_type() was pub but had no in-tree callers outside handle_find_information, so I went with the rename rather than keeping a misleadingly-named accessor around; happy to add get_att_type() alongside the old name instead if you’d prefer to avoid the technically-breaking change.

@davertay-j — thanks for the btmon capture and the hardware validation. You’re right that my PR description was wrong about BlueZ: it does not shrug off the reserved Format byte, it mis-parses the payload and can’t discover descriptors at all on a 128-bit service.  The 128-bit path here matches what you validated on the DA1469x; the 32-bit widening is spec-derived (same base-UUID expansion btuuid uses internally) and not yet on hardware, since a Uuid32 is hard to route through a real attribute table.
CI-equivalent checks pass locally: cargo +nightly fmt --check, clippy (gatt,peripheral,central,legacy-pairing), and the lib tests across all three CI feature sets.

alexmoon (approved):

lulf (approved):

PR #621, opened 2026-06-09
Configurable advertising packet filtering

stargazing-dino:
There's currently no way to specify whether we want to filter duplicate advertising packets.

```rs
// trouble/host/src/scan.rs:124
// The second true hardcodes it to deduping
host.command(LeSetScanEnable::new(true, true)).await?;
```

This PR just adds that field to `ScanConfig` and we thread it through.

I think this technically is a "breaking change" since we don't have a #[non_exhaustive] on ScanConfig.

But everywhere in the codebase uses `..Default::default()` as far as I can tell.

This will also be a small behavior change too for `scan` - it used to default to deduping - and now we unify on not deduping same as `scan_ext`. For consistency alone, I think this PR is worth it.

Legacy scan users who want back the old behavior, they'll need to opt back in:

```rs
ScanConfig {
  filter_duplicates: FilterDuplicates::Enabled
  ..
}
```

Let me know if you rather not the change of behavior, I think I might be able to finesse it but the diff will look a bit uglier

### DIFF
--- a/host/src/connection.rs
+++ b/host/src/connection.rs
@@ -7,8 +7,8 @@ use bt_hci::cmd::le::{
 use bt_hci::cmd::status::ReadRssi;
 use bt_hci::controller::{ControllerCmdAsync, ControllerCmdSync};
 use bt_hci::param::{
-    AllPhys, ConnHandle, DisconnectReason, FrameSpaceInitiator, LeConnRole, PhyKind, PhyMask, PhyOptions, SpacingTypes,
-    Status,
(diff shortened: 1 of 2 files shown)


lulf (approved):
Thank you!

augustoroman:
Is CI stuck? This is a convenient fix for me too (though I have pulled in the change locally).

lulf:
I'm moving the CI to another location, so will be done a little bit.

PR #624, opened 2026-06-13
Support concurrent connection + advertising

stargazing-dino:
Draft mode while I find some more boards to test on.

I attempted on esp32 but ran into an issue unrelated to this PR

### DIFF
--- a/examples/apps/src/ble_advertise_multiple.rs
+++ b/examples/apps/src/ble_advertise_multiple.rs
@@ -18,7 +18,7 @@ where
     let address: Address = Address::random([0xff, 0x8f, 0x1a, 0x05, 0xe4, 0xff]);
     info!("Our address = {:?}", address);
 
-    let mut resources: HostResources<_, DefaultPacketPool, 0, 0, 2> = HostResources::new();
+    let mut resources: HostResources<_, DefaultPacketPool, 1, 0, 3> = HostResources::new();
     let stack = trouble_host::new(controller, &mut resources)
         .set_random_address(address)
         .build();
@@ -28,8 +28,8 @@ where
     let mut adv_data = [0; 31];
     let len = AdStructure::encode_slice(
         &[
-            AdStructure::CompleteLocalName(b"Trouble Multiadv"),
             AdStructure::Flags(LE_GENERAL_DISCOVERABLE | BR_EDR_NOT_SUPPORTED),
+            AdStructure::CompleteLocalName(b"Trouble Multiadv"),
         ],
         &mut adv_data[..],
     )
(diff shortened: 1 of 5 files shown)


host/src/peripheral.rs
```diff
@@ -345,25 +354,62 @@ pub struct Advertiser<'d, C, P: PacketPool> {
...
+    /// Returns Error::Timeout if all sets stopped advertising, or Error::NotFound if none of the
+    /// sets is connectable.
+    pub async fn accept_connectable(&mut self) -> Result<Connection<'d, P>, BleHostError<C::Error>>
+    where
+        C: for<'t> ControllerCmdSync<LeSetExtAdvEnable<'t>>,
+    {
+        // Not setting `done`; Drop cancels advertising once the caller stops looping.
+        let set = self.connectable.ok_or(Error::NotFound)?;
+        let conn = self.accept_one().await?;
+        // Per BT Core Spec Vol 4, Part E, Section 7.8.56 the connectable set terminates right after
```
stargazing-dino, line 401: Verified against: [Core Specification 6.3](https://www.bluetooth.com/specifications/specs/core-specification-6-3/)

host/src/peripheral.rs
```diff
@@ -9,12 +9,15 @@ use bt_hci::cmd::le::{
...
 use bt_hci::param::{AddrKind, AdvChannelMap, AdvHandle, AdvKind, AdvSet, BdAddr, LeConnRole, Operation};
 use embassy_futures::select::{select, Either};
+use embassy_time::{with_timeout, Duration};
 
 use crate::advertise::{Advertisement, AdvertisementParameters, AdvertisementSet, RawAdvertisement};
 use crate::connection::Connection;
 use crate::host::BleHost;
 use crate::{bt_hci_duration, bt_hci_ext_duration, Address, BleHostError, Error, PacketPool};
 
+const REARM_TERMINATE_TIMEOUT: Duration = Duration::from_millis(100);
```
stargazing-dino, line 19: This value is arbitrary - just seemed reasonable. We can remove it, inline it or anything else if ya'll prefer.

examples/apps/src/ble_advertise_multiple.rs
```diff
@@ -28,8 +28,8 @@ where
     let mut adv_data = [0; 31];
     let len = AdStructure::encode_slice(
         &[
-            AdStructure::CompleteLocalName(b"Trouble Multiadv"),
             AdStructure::Flags(LE_GENERAL_DISCOVERABLE | BR_EDR_NOT_SUPPORTED),
+            AdStructure::CompleteLocalName(b"Trouble Multiadv"),
```
stargazing-dino, line 32: Running this example gave me a: `InvalidValue` (pre-existing issue it seems) - I guess the adv_data must start with the Flags

lulf (commented):
I'm not sure I understand the purpose. You can connect and advertise today if you spawn a task with the accepted handle, then advertise again. Though I agree it would be nice to return which AdvSet was matched.

stargazing-dino:
Oh interesting, yeah, you're right about passing it to a task and re-initializing - should totally work. I'll look into seeing if this covers my use-case and might close out this PR if so

stargazing-dino:
On second thought, I think this PR is the right direction but this changeset does too much. I'm going to break it up and better motivate upcoming ones.

PR #625, opened 2026-06-15
feat(security): Seed RNG from the controller, add additional crypto functions, and add bond info to Encrypted event

alexmoon:
This has several small improvements to the security manager:
- The RNG seed for the security manager is now fetched from the controller using the mandatory `LeRand` command. This eliminates the need for apps to provide an RNG to seed the security manager.
- Adds the missing `d1` diversifying crypto function to allow generating LTKs and IRKs from encryption/identity roots (respectively).
- Adds a `bond` field to the `ConnectionEvent::Encrypted` and `GattConnectionEvent::Encrypted` events which is set to `Some` when encryption is re-established with a stored bonded peer. This allows apps to differentiate between unbonded encryption events and bonded encryption events.

### DIFF
--- a/benchmarks/nrf-sdc/Cargo.lock
+++ b/benchmarks/nrf-sdc/Cargo.lock
@@ -13,9 +13,9 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "1.5.1"
+version = "1.5.0"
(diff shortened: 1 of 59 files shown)


lulf (approved):
Excellent!

PR #628, opened 2026-06-17
fix: L2CAP ACL fragmentation and ATT MTU exchange

alexmoon:
This fixes some issues with packet fragmentation:
- There were some edge cases around l2cap packet fragmentation where sizes were not being calculated correctly.
- There was no synchronization preventing other PDUs from interleaving with a fragmented PDU which would prevent successful reassembly.
- Large PDUs that exceeded the size of the controller buffers could not be sent.

I've reworked the fragmentation calculations. In an async context, we no longer require enough ACL buffers for the full PDU up front. Instead we acquire an `AclSendLock` to prevent other ACL packets from being sent then send ACL packets as buffers become available until the entire PDU is sent. This allows large PDUs to be sent even if the controller's buffers are small. Blocking sends still check for enough buffers up front because there's no opportunity for buffers to be returned in a sync context.

I've also included a change to prevent redundant ATT MTU exchanges after one succeeds.

### DIFF
--- a/host/src/channel_manager.rs
+++ b/host/src/channel_manager.rs
@@ -985,14 +985,14 @@ impl<'d, P: PacketPool> ChannelManager<'d, P> {
         let (first, remaining) = buf.split_at(buf.len().min(mps as usize - 2));
 
         let len = encode(first, &mut p_buf[..], peer_cid, Some(buf.len() as u16))?;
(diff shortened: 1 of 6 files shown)


lulf:
Looks good, but will run some more tests on this first.

lulf (approved):
Works in our tests, thanks!

PR #629, opened 2026-06-19
reproducer: No ACL reply from the ESP host (wokwi simulation)

Fristi:
## The problem

No ACL reply from the ESP host 
When BlueZ sends Read By Type for Server Supported Features (0x2b3a)

Time | Event | Meaning
-- | -- | --
08:07:23.519 | Disconnect handle 1 | Stale session from prior run
08:07:23.410 | Start Discovery | bluetoothctl scan
08:07:23.525+ | ADV from 9A:78:56:34:12:02 | Peripheral visible, UUID 0xaa8a
08:07:31.565 | LE Create Connection | BlueZ connects
08:07:31.822 | LE Connection Complete | Link up
08:07:31.836–849 | Exchange MTU 517→251 | Works (handled inline in trouble-host RX)
08:07:31.850 | Read By Type 0x2b3a | No response
08:07:35.614 | LE Create Connection Cancel | BlueZ gives up

This is the same thing I'm seeing with a central program, written in Rust but with btleplug acting as a central role

**FYI:** it's a reproducer of a issue I'm seeing with this setup (might be a potential issue in trouble?). The simulator might be of interest to have in the project as an example, but if it's not the right place for this I fully understand :) No stress to merge this in, but having the reproducer here, might make it easier to debug straight with the source

## Background

I've been exploring an approach where I can test my firmware in a simulator. The reason to have e2e test, which uses the firmware in a simulator and talks over UART instead of radio so the central node (which is linux/bluez program) can talk directly to the firmware. If there other approaches to this, I'm keen on hearing and it would be nice to document this :)

I came across wokwi, which can talk i2c, uart, pwm and so on. Since there is work being done in bt-hci on SerialTransport, I thought I could have wokwi talk BLE over UART. 

Turns out UART0 is exposed over TCP, but is used for logging usually (that's the reason I've turned that off). I've requested to have another TCP port being exposed for UART1 at wokwi, but that's not in yet.

I have some experience with BLE, Rust and esp32, but it's not extensive as others here. So I've been using Cursor as an aid in this endeavour. 

Architecture:
  Wokwi ESP32-C3 (ble_bas_peripheral_hci)
    → UART0 GPIO21/20 @ 115200 HCI H4
    → Wokwi RFC2217 :4000
    → socat PTY /tmp/trouble-hci
    → edge-hci-bridge.py (dual Bumble controllers on LocalLink)
    → BlueZ vhci → bluetoothctl / btmon

In the README there is a playbook on how to run this wokwi simulation. 

First I was using the 0.6, but then Cursor copied the whole source of trouble and fixed this: https://github.com/embassy-rs/trouble/commit/ea3872b7dc29995b7a083b57ed205be195c653ac, so I switched to the latest git revision for now


lulf (commented):
Can you add this to ci.sh? I don't want to merge things that can bitrot.

Fristi:
Agreed! I will look into this!

Fristi:
Closing this for now, I think this approach could be worthwhile in a separate repository to avoid too much integrations ending up here. Also when wokiwi has UART2 support and some more work is done on trouble, this is something to pursue

PR #631, opened 2026-06-21
fix(gap): store device name by reference instead of a one-shot StaticCell

TuEmb:
Device name is a caller-provided `&'a str` that already outlives the attribute table, so it can be stored by reference; the one-shot StaticCell copy was unnecessary and made the GAP service un-rebuildable.

```
running 1 test
test gap::tests::gap_config_can_be_built_twice ... FAILED

failures:

---- gap::tests::gap_config_can_be_built_twice stdout ----

thread 'gap::tests::gap_config_can_be_built_twice' (60124) panicked at C:\Users\ION\.cargo\registry\src\index.crates.io-1949cf8c6b5b557f\static_cell-2.1.1\src\lib.rs:84:13:
`StaticCell` is already full, it can't be initialized twice.


failures:
    gap::tests::gap_config_can_be_built_twice

test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 22 filtered out; finished in 0.00s

```

### DIFF
--- a/host/src/gap.rs
+++ b/host/src/gap.rs
@@ -127,3 +127,27 @@ impl<'a> CentralConfig<'a> {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod tests {
+    extern crate std;
+
+    use embassy_sync::blocking_mutex::raw::NoopRawMutex;
+
+    use super::*;
+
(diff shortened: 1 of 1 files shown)


lulf (approved):
Thanks

PR #632, opened 2026-06-22
fix: handle disconnected links and channels gracefully

lulf:
Ensure that polling an already disconnected link is not causing the adapter to fail.


lulf:
bender run

PR #634, opened 2026-06-28
Update to next bt-hci

bugadani:
This PR test-runs the alloc_buf API in the current git version of bt-hci.


lulf:
Hi there! I haven't kept close track on the state of this, but I see there are some build failures. Is there something you need me to release or just... do to move it forward?

bugadani:
Eventually, bt-hci will need to be released, and HALs need to be adjusted before this can progress. I've got his PR to this point to see if the driver crate is implementable in bt-hci-linux and -usb. I guess we'll just need some time for the rest.

PR #637, opened 2026-07-03
fix(perf)!: avoid recomputing local public key in dh_key

joelsa:
`SecretKey::dh_key()` recomputes the local public key from the secret scalar (`p256::PublicKey::from_secret_scalar`) solely to implement a simple check (reject a peer public key that is invalid or identical to our own). 

I found that on my `nRF52810` target this costs quite some time and more importantly also some stack usage, and I am really limited there with only 24K.

The recomputation is actually unnecessary: both call sites (central and peripheral pairing) already hold the local public key, which is computed once, cached in the security manager, and transmitted to the peer in the same procedure.

This is a breaking change to a public API however. If you'd rather keep `dh_key(pk)` stable, I'm happy to rework this as an internal method used by the pairing code, and keep the old signature as a wrapper then.

Also: Is there some interest in me contributing my `nRF52810` examples or rather not (in order to not hint at any support, the 24K RAM really is limiting)?

### DIFF
--- a/host/src/security_manager/crypto.rs
+++ b/host/src/security_manager/crypto.rs
@@ -411,8 +411,11 @@ impl SecretKey {
     /// Computes a shared secret from the local secret key and remote public
     /// key. Returns [`None`] if the public key is either invalid or derived
     /// from the same secret key ([Vol 3] Part H, Section 2.3.5.6.1).
+    ///
+    /// `local_pk` must be the public key of `self`; passing it in avoids an
+    /// expensive scalar multiplication to recompute it.
     #[must_use]
-    pub fn dh_key(&self, pk: PublicKey) -> Option<DHKey> {
+    pub fn dh_key(&self, pk: PublicKey, local_pk: &PublicKey) -> Option<DHKey> {
         use p256::elliptic_curve::sec1::FromEncodedPoint;
         if pk.is_debug() {
             return None; // TODO: Compile-time option for debug-only mode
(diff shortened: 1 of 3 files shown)


lulf (commented):
Examples should demonstrate some feature or concept, but not necessarily covering "everything". Not having looked at your examples, I'm not sure if they would fit in or not. But in general having a lot of examples that should be tested and verified causes some maintenance overhead, so it really depends on whether on how much useful it is in general to users.

joelsa:
> so it really depends on whether on how much useful it is in general to users.

I would say it is useful if there are users that want to use an nRF52810, because then having the memory.x and everything in place already will save them some time figuring it out by themselves.

But I feel like trouble is not really suited for that particular MCU anyway, because even for some simple pairing I had to resort to `DEFMT_LOG = INFO` and

```toml
opt-level = "z"
lto = true
codegen-units = 1
```
just so it (barely) fits inside the 192K flash and 24K ram.

So if I am honest I would tell anyone who has to use the 52810 for cost/legacy reasons to use the S112 softdevice.

lulf:
Point taken, though I think the defaults are neither flash nor ram optimized, so if you can improve that, it would be helpful.  There are lots of knobs, some are covered by https://embassy.dev/trouble/#_performance - but you still need to know where to look I suppose.

In my experience, for l2cap, the flash and RAM size in the firmware at work is lower than with the legacy softdevice, but I have not measured/tuned the efficiency of GATT so far, so there could be some room for improvement there.

lulf (approved):

PR #638, opened 2026-07-03
feat(central): Various improvements to central

alexmoon:
This PR has a handful of improvements to the central and scan modules.

Scanning:
- `ScanSession` is now a future that resolves when the scan times out (if the scan config has a timeout)
- `ScanSession::stop()` is a method to stop the scan then asynchronously wait for the scan to be cancelled in the controller. This is needed to know when it is safe to try to use the central for something else, such as connecting to a discovered peer.
- `Scanner` now mutably borrows from `Central` instead of taking ownership. This makes it much easier to deal with lifetimes for applications that alternately scan and connect (no more `into_inner()` and having to pass the `Central` around by value everywhere).

The filter accept list is a shared resource (between advertising, scanning and connecting) on the controller so I've made several changes to tighten up when it is used to reduce contention:
- If a scan config has an empty filter accept list, the filter accept list is not touched.
- When connecting to a single peer, a direct connection command is used instead of the filter accept list.

The change to using direct connection commands exposed an issue with using the controller generated resolvable private addresses. Specifically, the controller will only use the generated RPA when the peer address of the command can be found in the resolving list. There are a few cases where this can fail:
- Active scanning of an unbonded peer (pre-existing issue)
- Initiating direct connection to an unbonded peer (new issue due to the filter accept list changes)
- Directed advertising to an unbonded peer (pre-existing issue)

I've added functionality to the host to generate RPAs when the `central` and `security` features are both enabled. This resolves the first two issues.

Directed advertising to an unbonded peer using legacy advertising (e.g. `advertise`) will use RPAs when the `central` feature is enabled, but will still use the identity address when the `central` feature is not enabled. Directed advertising to an unbonded peer using extended advertising (e.g. `advertise_ext`) will use the identity address unless an alternative random address is provided as the `address` in the `AdvertisementSet`. My belief is that directed advertising to unbonded peers should be quite rare and since it can be worked around by the application using extended advertising and providing an address in the advertisement set I kept my changes restricted to the `central` feature flag. Let me know if you'd prefer I remove that limitation.

### DIFF
--- a/examples/apps/src/ble_scanner.rs
+++ b/examples/apps/src/ble_scanner.rs
@@ -24,13 +24,13 @@ where
     let stack = trouble_host::new(controller, &mut resources)
         .set_random_address(address)
         .build();
-    let central = stack.central();
+    let mut central = stack.central();
     let mut runner = stack.runner();
 
     let printer = Printer {
         seen: RefCell::new(Deque::new()),
     };
-    let mut scanner = Scanner::new(central);
+    let mut scanner = Scanner::new(&mut central);
     let _ = join(runner.run_with_handler(&printer), async {
         let mut config = ScanConfig::default();
         config.active = true;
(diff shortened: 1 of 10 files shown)


lulf (approved):
Thanks!

PR #639, opened 2026-07-07
feat: add HCI ISO (CIS) event and data support

LegitCamper:
Adds a Stack::iso() handle for isochronous-stream HCI commands and data, dispatches LE CIS Request/Established events and incoming ISO data packets to EventHandler (previously silently dropped by RxRunner's catch-all arms), and enables the corresponding event mask bits so controllers actually emit them.


lulf (commented):
Thanks! Could you please guard this behind a cargo feature flag? To prevent additional flash usage for firmware that does not need to support iso.

lulf (approved):
Thanks!

PR #640, opened 2026-07-09
docs: update HostResources generics in getting started guide

constkarma:
The getting started guide still documents the old `HostResources` generic parameter list.

This updates the `HostResources` section to include the controller type parameter and the bond storage parameter used by the current API. It also updates the example generic parameter order and notes that the controller type can often be inferred when the resources are passed to `trouble_host::new`.

- https://docs.rs/trouble-host/0.7.0/trouble_host/struct.HostResources.html
- https://docs.rs/trouble-host/0.6.0/trouble_host/struct.HostResources.html


lulf (approved):
Thanks!

PR #641, opened 2026-07-10
feat: add security-p256-cortex-m4 feature

joelsa:
Use the assembly-optimized `p256-cortex-m4 `crate (https://crates.io/crates/p256_cortex_m4) for LESC pairing on Cortex-M4 targets. Roughly 10x faster and far less stack usage than the pure-Rust p256 crate, which matters for my nRF52810 target.

I added a benchmark (not on this branch) here: https://github.com/joelsa/trouble-fork/tree/p256-cortex-m4-benchmark

The results are:

```
[INFO ] P-256 LESC pairing microbenchmark (8 iterations) (p256_bench src/bin/p256_bench.rs:193)
[INFO ] TIMER0 @ 16 MHz --- 1 tick = 62.5 ns = 4 CPU cycles @ 64 MHz (p256_bench src/bin/p256_bench.rs:194)
[INFO ] p256 (pure Rust, `security` feature): (p256_bench src/bin/p256_bench.rs:196)
[INFO ]     keygen: avg 1437 us (92004 cyc)  min 1426 us  max 1452 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ]     public_key: avg 155944 us (9980432 cyc)  min 155944 us  max 155944 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ]     ecdh: avg 155927 us (9979328 cyc)  min 155927 us  max 155927 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ]     pairing: avg 313308 us (20051764 cyc)  min 313297 us  max 313323 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ] p256-cortex-m4 (asm, `security-p256-cortex-m4` feature): (p256_bench src/bin/p256_bench.rs:203)
[INFO ]     keygen: avg 1420 us (90880 cyc)  min 1400 us  max 1438 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ]     public_key: avg 7648 us (489528 cyc)  min 7648 us  max 7648 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ]     ecdh: avg 21834 us (1397432 cyc)  min 21834 us  max 21834 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ]     pairing: avg 30903 us (1977840 cyc)  min 30884 us  max 30922 us (p256_bench src/bin/p256_bench.rs:76)
[INFO ] Speedup (cortex-m4 vs pure Rust): (p256_bench src/bin/p256_bench.rs:210)
[INFO ]     keygen: 1.0x faster (p256_bench src/bin/p256_bench.rs:175)
[INFO ]     public_key: 20.3x faster (p256_bench src/bin/p256_bench.rs:175)
[INFO ]     ecdh: 7.1x faster (p256_bench src/bin/p256_bench.rs:175)
[INFO ]     pairing: 10.1x faster (p256_bench src/bin/p256_bench.rs:175)
[INFO ] done (p256_bench src/bin/p256_bench.rs:216)
```

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -33,6 +33,7 @@ futures = { version = "0.3", default-features = false }
 heapless = "0.9"
 trouble-host-macros = { path = "../host-macros", version = "0.5.0", optional = true }
 p256 = { version = "0.13.2", default-features = false, features = ["ecdh","arithmetic"], optional = true }
+p256-cortex-m4 = { version = "0.1.0-alpha.6", default-features = false, optional = true }
 rand_core = { version = "0.6", optional = true }
 rand_chacha = { version = "0.3", default-features = false, optional = true }
 rand = { version="0.8", default-features = false, optional = true}
@@ -82,6 +83,8 @@ connection-metrics = []
 # Enable additional channel metrics
 channel-metrics = []
 security = [ "dep:p256", "dep:aes", "dep:cmac", "dep:rand_core", "dep:rand_chacha", "dep:rand" ]
+# Use the assembly-optimized P-256 implementation for LESC pairing (ARM Cortex-M4 targets only).
+security-p256-cortex-m4 = [ "security", "dep:p256-cortex-m4" ]
 # Enable LE Legacy Pairing (BLE 4.0/4.1 fallback when peer doesn't support Secure Connections)
 legacy-pairing = ["security"]
 
--- a/host/src/host.rs
+++ b/host/src/host.rs
@@ -44,9 +44,9 @@ use bt_hci::param::{
 };
 use bt_hci::{ControllerToHostPacket, FromHciBytes, WriteHci};
 use embassy_futures::select::{select3, select5, Either3, Either5};
-#[cfg(any(feature = "scan", all(feature = "security", feature = "central")))]
+#[cfg(any(feature = "scan", feature = "security"))]
 use embassy_sync::blocking_mutex::raw::NoopRawMutex;
-#[cfg(all(feature = "security", feature = "central"))]
+#[cfg(feature = "security")]
 use embassy_sync::mutex::Mutex;
 use embassy_sync::once_lock::OnceLock;
 #[cfg(feature = "scan")]
--- a/host/src/security_manager/crypto.rs
+++ b/host/src/security_manager/crypto.rs
@@ -7,6 +7,7 @@ use aes::cipher::{BlockEncrypt, KeyInit};
 use aes::Aes128;
 use bt_hci::param::BdAddr;
 use cmac::digest;
+#[cfg(not(feature = "security-p256-cortex-m4"))]
 use p256::ecdh;
 use rand_core::{CryptoRng, RngCore};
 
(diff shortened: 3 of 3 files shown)


xoviat:
stm32 has hardware accelerators for this. I wonder whether this could be exposed as a generic trait or something.

joelsa:
That's a phenomenal idea!

AFAIK, embedded-cal is WIP, but `p256.rs` was added three weeks ago:
https://github.com/lake-rs/embedded-cal/blob/main/embedded-cal/src/p256.rs

joelsa:
@chrysn: Do you think a generic `p256` trait would be in-scope for embedded-cal? Is this something I could help with?

xoviat:
is this trait not sufficient? 

https://github.com/lake-rs/embedded-cal/blob/main/embedded-cal/src/dh.rs#L27

joelsa:
Good question.

I'd say when using hardware accelerators some of the functions should become fallible, also I am unsure if every hardware accelerator has a way to fulfill `generate_visible`. And for trouble as a consumer, an async version would probably be preferable?

So I'd say it is _much more_ than sufficient for our use case, but might put too much burden on impls.

chrysn:
Thanks for pinging me!

I'd have to read up on what Bluetooth needs, but if it's ECDH, then yes that trait should be sufficient.

With respect to hardware generators and fallibility, we currently treat the hardware accelerators as part of the hardware, and let things crash just like things would crash if a GPIO or UART write or read suddenly caused a bus fault. (And soliciting feedback on that in https://github.com/lake-rs/embedded-cal/issues/46).

There will always be a `generate_visible` – if the accelerator doesn't have a means to generate a key with the option to export it, then the software layers on top will generate a random key and import it into the secure area. But you'd only need `generate_visible` for persisted private keys, which from a glimpse at the current `host/src/security_manager/crypto.rs` you don't need.

As for async, the line of thinking in embedded-cal is that it's replacing software implementations (which are blocking unless deferred to a 2nd core), so any acceleration would be faster than what users are ready to block on, and don't expect it to be worth triggering the async machine. (But there too, feedback welcome).

Coming to implementation burden: Not sure what you mean precisely, but embedded-cal does not have a stable release yet, so it's still brittle to build on. We provide embedded-cal implementations for some platforms (including stm32wb55a), and some backed by rustcrypto and libcrux. Looking at the numbers in the original post, we should probably add https://crates.io/crates/p256_cortex_m4 support as well (PRs welcome, as we're currently all busy reviewing each other's PRs).

joelsa:
Thank you very much for this detailed answer!

The one concrete example I had in mind is the crypto companion IC Microchip ATECC608A, which I've seen used in quite some projects now. It allows to "lock away" a private key, which is then not extractable anymore, but it can still solve challenges. It supports ECC P256 (ECDH and ECDSA), SHA256 and AES-GCM.

This then produces all of the problems mentioned above. It is connected via I2C or Single-Wire, so the bus can fail in significantly more ways and more often than an internal bus in a MCU and the I2C transactions can also take several ms.

But maybe this IC sits in a weird spot between a HSM/TPM and a classical built-in accelerator and is niche enough so that it should not influence decisions.

joelsa:
Thanks for merging!

Picking up the discussion from the embassy chat, I'll leave this as a pointer here: https://github.com/xoviat/global-crypto
Thanks @xoviat for sketching this

HaoboGu:
Adding this feature introduces compile-time dependency of `arm-none-eabi-gcc` binary. Is it possible to add a full Rust implementation using same asm code? I checked the source, it seems quite feasible

joelsa:
Yes, this absolutely is possible, but as this README says: https://crates.io/crates/p256-cm4 it would need conditional compilation to ensure interop

PR #642, opened 2026-07-11
fix(gatt): clear prepare-write queue on a rejected Execute Write

stephencox-inveya:
## Problem

When a queued (prepare/execute) write is **rejected** by the application via `WriteEvent::reject`, the prepare-write queue is left populated. The next client's Prepare Write then fails the offset-contiguity check in `PrepareWriteState::queue` (`self.offset + self.len != offset`) and returns `PREPARE_QUEUE_FULL` — so on a peripheral that rejects some long writes (e.g. an authorization gate), **every other rejected long write fails**.

`process_accept` clears the queue via the attribute server's `handle_execute_write`, and `accept_unprocessed` clears it explicitly, but `process_reject` never did.

Per Core Spec Vol 3, Part F, §3.4.6.3, an Execute Write empties the prepare queue regardless of outcome.

## Fix

In `process_reject`, when the rejected request is an `ExecuteWrite`, call `connection.clear_prepare_write()` (gated on `att-queued-writes`, mirroring `accept_unprocessed`).

## Test

Adds `gatt::tests::test_execute_write_reject_clears_prepare_queue`: queue a prepared write, reject its Execute Write, assert the queue is empty. Fails without the fix (`rejected Execute Write must clear the prepare queue`), passes with it.

Found while bench-testing a peripheral whose app rejects unauthorized long writes: consecutive rejected writes alternated `ok / PREPARE_QUEUE_FULL`; verified fixed on hardware (nRF52833).

### DIFF
--- a/host/src/gatt.rs
+++ b/host/src/gatt.rs
@@ -830,6 +830,16 @@ fn process_reject<'stack, P: PacketPool>(
     let Att::Client(att) = att else {
         unreachable!("Expected Att::Client, got {:?}", att)
     };
+    // An Execute Write empties the prepare queue regardless of outcome (Core
+    // Spec Vol 3, Part F, §3.4.6.3). `process_accept` clears it via the attribute
+    // server (`handle_execute_write`), and `accept_unprocessed` clears it too, but
+    // the reject path did not — leaving a stale queue that makes the *next*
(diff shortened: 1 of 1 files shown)


lulf (approved):
Looks ok, thanks! Personally I think the comments are a bit on the verbose side.

PR #643, opened 2026-07-12
disable services at compile time via conditional compilation attributes

ajwahab:
For example, `BatteryService` may be omitted from the GATT Server by disabling the "ble-bas" feature.

```rust
#[gatt_server]
struct MyGattServer {
    hrs: HeartRateService,
    #[cfg(feature = "ble-bas")]
    bas: BatteryService,
}
```


lulf (approved):
Nice improvement, thanks!

PR #648, opened 2026-07-19
fix(esp32): pass SoftwareInterrupt to esp_rtos::start on Xtensa

KarpagamKarthikeyan:
## Summary

The ESP32 examples currently gate `SoftwareInterrupt` creation and the second argument to `esp_rtos::start()` behind `#[cfg(target_arch = "riscv32")]`:

```rust
esp_rtos::start(
    timg0.timer0,
    #[cfg(target_arch = "riscv32")]
    software_interrupt.software_interrupt0,
);
```

However, with the currently pinned `esp-hal` revision (`b7eec0f`), `esp_rtos::start()` requires the software interrupt argument on **all** architectures. As a result, the second argument is omitted on Xtensa targets (`esp32`, `esp32s3`), causing the examples to fail to compile with:

```text
error[E0061]: this function takes 2 arguments but 1 argument was supplied
```

This issue is not caught by CI because it currently only builds the RISC-V (`esp32c3`) examples.

This PR removes the `riscv32`-only gating across all nine example binaries so that the software interrupt is always created and passed to `esp_rtos::start()`.

## Testing

* ✅ Builds and links successfully for:
  * `esp32` (`xtensa-esp32-none-elf`)
  * `esp32s3` (`xtensa-esp32s3-none-elf`)
  
* ✅ Flashed `ble_bas_peripheral` to real ESP32 hardware:
  * Device advertises correctly
  * Accepts BLE connections
  * Sends GATT notifications successfully
  
* ✅ Verified no regressions on RISC-V:
  * `esp32c3` (`riscv32imc`)
  * `esp32c6` (`riscv32imac`)

## Out of Scope

The `ble_bas_peripheral_bonding --features security` example currently fails to build on both Xtensa and RISC-V for an unrelated reason. The `examples/esp32` crate does not declare `embedded-storage` as a direct dependency (it is only pulled in transitively via `esp-storage`), so the `use embedded_storage::...` import fails and `capacity()` resolves to a private field instead of the `ReadNorFlash` trait method. CI does not catch this because the bonding example requires the `security` feature, which is not built for the ESP32 target. This can be addressed separately by adding `embedded-storage = "0.3.1"` to the example's `Cargo.toml`.

### DIFF
--- a/examples/esp32/src/bin/ble_bas_central.rs
+++ b/examples/esp32/src/bin/ble_bas_central.rs
@@ -18,14 +18,9 @@ async fn main(_s: Spawner) {
     esp_alloc::heap_allocator!(size: 72 * 1024);
     let timg0 = TimerGroup::new(peripherals.TIMG0);
 
-    #[cfg(target_arch = "riscv32")]
     let software_interrupt = esp_hal::interrupt::software::SoftwareInterruptControl::new(peripherals.SW_INTERRUPT);
 
-    esp_rtos::start(
-        timg0.timer0,
-        #[cfg(target_arch = "riscv32")]
-        software_interrupt.software_interrupt0,
-    );
+    esp_rtos::start(timg0.timer0, software_interrupt.software_interrupt0);
 
     let bluetooth = peripherals.BT;
     let connector = BleConnector::new(bluetooth, Default::default()).unwrap();
--- a/examples/esp32/src/bin/ble_bas_central_multiple.rs
+++ b/examples/esp32/src/bin/ble_bas_central_multiple.rs
@@ -18,14 +18,9 @@ async fn main(_s: Spawner) {
     esp_alloc::heap_allocator!(size: 72 * 1024);
     let timg0 = TimerGroup::new(peripherals.TIMG0);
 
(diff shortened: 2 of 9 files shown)


KarpagamKarthikeyan:
Gentle ping on this — CI is green and it's a small, focused change (removing a `#[cfg(target_arch = "riscv32")]` gate in 9 example bins). It repairs the esp32/esp32s3 (Xtensa) example build, which currently fails because `esp_rtos::start` now requires the `SoftwareInterrupt` argument on all architectures. Verified on real esp32 hardware — boots and advertises. Would appreciate a review when someone has a moment. Thanks!

cc @lulf @HaoboGu

HaoboGu (approved):
lgtm

KarpagamKarthikeyan:
Gentle ping on this —  Thanks again for the review and approval. Since the PR has an approving review and the checks have passed, I was wondering what the remaining steps are to get this merged. Is there anything else you'd like me to address, or is it just waiting for a maintainer to merge? Thanks!

cc @lulf @HaoboGu

HaoboGu:
Thanks!

PR #652, opened 2026-07-23
feat: expose an event-driven host startup readiness wait

gudipudiramanakumar:
## Summary

Add `Stack::wait_ready()` so applications can await completion of the `ControlRunner` startup sequence before starting role operations such as advertising, scanning, or connecting.

This avoids application-side timing gates. In particular, it allows an application to run the host runner and a role task concurrently without letting advertising (or other HCI commands) contend with the host's initial command sequence.

The readiness notification is emitted only after controller startup commands, the device-address read, and optional privacy initialization complete.

## Motivation

The host already tracks initialization internally via `HostState::initialized`, a `OnceLock<InitialState>` that is populated partway through `ControlRunner` startup. That lock is crate-private and is intended for internal use (for example, gating `BleHost::command()` and the TX runner), not as an application-facing readiness API.

It is also the wrong synchronization point for role tasks:

- `initialized` is set before the address read and optional privacy setup finish.
- `OnceLock::get().await` blocks until that earlier milestone, not until the host is fully ready for role operations.
- It offers no public, event-driven way for application tasks to coordinate with startup completion.

Without a dedicated readiness API, applications have no reliable way to know when startup is done and have resorted to fixed delays. `Stack::wait_ready()` exposes an explicit, signal-based synchronization point that matches when `ControlRunner` actually finishes startup.

## Example

```rust
let runner = stack.runner();
let mut peripheral = stack.peripheral();

join(
    runner.run(),
    async {
        stack.wait_ready().await;
        peripheral.advertise(&Default::default(), advertisement).await?;
        Ok::<(), BleHostError<_>>(())
    },
)
.await;
```

### DIFF
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -845,6 +845,9 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
     /// Wait for the host runner to complete controller startup before starting
     /// a role operation such as advertising, scanning, or connecting.
     pub async fn wait_ready(&self) {
+        if self.host_state.ready.signaled() {
+            return;
+        }
         self.host_state.ready.wait().await;
     }
 
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -845,10 +845,7 @@ impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
     /// Wait for the host runner to complete controller startup before starting
     /// a role operation such as advertising, scanning, or connecting.
     pub async fn wait_ready(&self) {
-        if self.host_state.ready.signaled() {
-            return;
-        }
-        self.host_state.ready.wait().await;
+        let _ = self.host_state.ready.get().await;
     }
 
     /// Obtain a [`Runner`] to drive the BLE host.
--- a/host/src/lib.rs
+++ b/host/src/lib.rs
@@ -842,6 +842,12 @@ impl<'stack, C, P: PacketPool> Stack<'stack, C, P> {
 }
 
 impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
+    /// Wait for the host runner to complete controller startup before starting
+    /// a role operation such as advertising, scanning, or connecting.
+    pub async fn wait_ready(&self) {
+        self.host_state.ready.wait().await;
+    }
+
     /// Obtain a [`Runner`] to drive the BLE host.
     ///
     /// The runner must be polled (e.g. via [`Runner::run()`]) to drive the BLE host.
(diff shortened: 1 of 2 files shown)


HaoboGu:
I wanted it too, but can you check whether the `NoopRawMutex` works instead of `CriticalSectionRawMutex`?

host/src/lib.rs
```diff
@@ -842,6 +842,12 @@ impl<'stack, C, P: PacketPool> Stack<'stack, C, P> {
 }
 
 impl<'stack, C: Controller, P: PacketPool> Stack<'stack, C, P> {
+    /// Wait for the host runner to complete controller startup before starting
+    /// a role operation such as advertising, scanning, or connecting.
+    pub async fn wait_ready(&self) {
```
HaoboGu, line 847: The second call of the `wait_ready` hangs forever.

gudipudiramanakumar: Yeah fixed that, and also modified to use NoopRawMutex

HaoboGu: I don't think it will work.

HaoboGu: `wait()` consumes the signal

gudipudiramanakumar: Yeah got it, can I use OnceLock?

HaoboGu: Is it no_std compatible?

gudipudiramanakumar: Yes `embassy_sync::OnceLock` is `no_std`. It is already used for `HostState::initialized`, `ready` is a second `OnceLock<()>` that fires after the full startup sequence. `cargo check` passes on embedded targets.

HaoboGu: That's great

lulf:
I'm not sure I think this is the right fix... I think peripheral.advertise() should just deal with/wait for the host to run (which I thought it already does) instead of the application needing to do that

gudipudiramanakumar:
> I'm not sure I think this is the right fix... I think peripheral.advertise() should just deal with/wait for the host to run (which I thought it already does) instead of the application needing to do that

Agreed, the application shouldn't have to do this, I'll move the wait inside the role operations and drop `Stack::wait_ready()`.

It does already wait, but on `initialized`, which `ControlRunner::run()` sets partway through startup (host.rs:1558) — before `ReadBdAddr` and, with `security`, the RPA timeout and initial resolving-list sync. So advertising unblocks while the runner is still issuing startup commands. On a controller reporting a single HCI command credit that stops the peripheral coming up, waiting until startup finishes fixes it. Reproduced on an EFR32MG24 running trouble-host over Zephyr's HCI RAW channel.

PR #655, opened 2026-07-29
wip: impl. embedded-cal

xoviat:
implements optional hardware acceleration with embedded-cal. the dhkey probably needs to be threaded through with a mutex.


xoviat:
bender run

PR #659, opened 2026-07-31
Add ability to provide a static passkey

### LINKED ISSUE #653: Provide the passkey directly instead of generating a random one
Currently, (for example for `IoCapabilities::DisplayOnly`) the passkey is generated as a random number here:
https://github.com/embassy-rs/trouble/blob/c21b1239fb86de712fc99da803def88378d81177/host/src/security_manager/pairing/peripheral.rs#L309

My question is, would a pull request to pass the passkey directly (optionally of course, maybe behind a feature) be accepted? While it's not an official documented BLE spec, some devices on the market already do such things (On a label on that device)

themkoi:
Closes #653.

Implemented the option for users to supply their own custom passkey instead of relying solely on the randomly generated one. If no custom passkey is provided, it falls back to the default random generator.

If you are willing to merge then suggest any feedback or changes that are needed.


host/src/security_manager/pairing/peripheral.rs
```diff
@@ -306,11 +306,13 @@ impl Pairing {
             }
             PairingMethod::PassKeyEntry { peripheral, .. } => {
                 if peripheral == PassKeyEntryAction::Display {
-                    phase_data.local_secret_rb = rng.sample(rand::distributions::Uniform::new_inclusive(0, 999999));
+                    // Use configured passkey if available, otherwise generate a random one
+                    let passkey = ops
+                        .passkey()
+                        .unwrap_or(rng.sample(rand::distributions::Uniform::new_inclusive(0, 999999)));
```
Dominaezzz: [`unwrap_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.unwrap_or_else) instead?

themkoi: changed it to unwrap_or_else any other changes I should make?

lulf (commented):
Clicked merge before I thought about it... this should definitely be guarded behind a feature that indicates this should only be used for testing.

Dominaezzz:
Why only for testing?

lulf:
> Why only for testing?

Well, I just want to make sure users understand that they should use a secure rng to generate this, not just hand in the key 1234?

lulf (approved):
Thinking more about this, I think the current approach is ok.

themkoi:
Resolved the merge conflict and fixed checks, @lulf  can we merge now?

PR #664, opened 2026-08-10
fix(connection): wait for the real LeConnectionUpdateComplete status

renkenono:
Previously, `Connection::update_connection_params()` was not awaiting for the LL procedure's outcome, which in cases where `UNSUPPORTED_REMOTE_FEATURE` was returned, the connection would timeout within 1-17 seconds in practice (* See Testing). This is because `trouble` does not detect the latter and therefore does not perform the L2CAP fallback logic. BlueZ parameters are probably too strict for the peripheral I am using.

With this change, the L2CAP fallback would happen and BlueZ would end up respecting it eliminating `Connection Timeout` incidents.

### Testing

- Peripheral: ESP32-C6 running `esp-hal` a patched variant of version 1.1.0.
- Central: GNU/Linux `6.17.0-124040-tuxedo #40~24.04.1tux1`, BlueZ 5.72, with a MediaTek MT7921/MT922 (`mt7921e`) onboard Bluetooth controller which returns `Unsupported Remote Feature` for the LL procedure.

Caveat: this PR is based on `main` which bumped `bt-hci` to 0.9. Since `esp-hal` doesn't support that version yet, I backported my fix to an older version and validated my changes against that instead. It would be nice if `trouble` and `esp-hal` play nicely to simplify the integration of `bt-hci` updates.

### DIFF
--- a/host/src/channel_manager.rs
+++ b/host/src/channel_manager.rs
@@ -722,10 +722,14 @@ impl<'d, P: PacketPool> ChannelManager<'d, P> {
                 }
                 L2capSignalCode::CONN_PARAM_UPDATE_RES => {
                     let res = ConnParamUpdateRes::from_hci_bytes_complete(signal_data)?;
-                    debug!(
-                        "[l2cap][conn = {:?}] connection param update response: {}",
-                        conn, res.result,
-                    );
+                    if res.result == 0 {
+                        debug!("[l2cap][conn = {:?}] connection param update accepted", conn);
+                    } else {
+                        warn!(
+                            "[l2cap][conn = {:?}] connection param update rejected result={}",
+                            conn, res.result
+                        );
+                    }
                     Ok(())
                 }
                 _ => {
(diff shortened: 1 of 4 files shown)


lulf:
I acknowledge the problem, but I don't think this is the right fix. It is not the intention for these methods to wait for the events, as those are instead exposed out to the application using the connection events. 

Instead, I think maybe a fix could be to use the https://www.bluetooth.com/wp-content/uploads/Files/Specification/HTML/Core-54/out/en/host-controller-interface/host-controller-interface-functional-specification.html#UUID-5cef6d01-c629-c784-03a5-490eb9b0408e in the case where the role is a peripheral, to perform an l2cap fallback before issuing the command.

renkenono:
Understood, yeah, I'm not too familiar with the protocol which is why I went for the straightforward solution that was visible from reading the code alone. I'll have to check how can I do that using the `stack` variable inside the `update_connection_params` method then it should be relatively more straightforward. Perhaps you have other pointers on how to implement this? Do we already have a copy of the remote features somewhere? Should we cache it or request it on demand?

lulf:
Try importing and using the LeReadRemoteFeatures command from bt-hci https://github.com/embassy-rs/bt-hci/blob/main/bt-hci/src/cmd/le.rs#L232 - I think start with request on demand, let's see how that performs first.

PR #665, opened 2026-08-14
Put necessary extra gating, make sure `peripheral` does not drag the multirole SDC lib

ivmarkov:
Background:
Matter Accessories need only the Peripheral role, and can get away with legacy advertising.
This means, I can link-in the much smaller `libsoftdevice_controller_peripheral.a` (which is the Peripheral one and which does NOT contain extended advertising) rather than pulling the Multirole one - which - besides peripheral + central also contains the ext adv code).

I think that was the case anyway before Extended Advertising came to be recently, i.e. this is a flash size regression.

What the PR introduces:
- A new `extended-advertising` feature, so that `peripheral` means the much slimmer (in terms of Flash) Peripheral + Legacy-only advertising
- A bunch of gates, not just for the new `extended-advertising` feature, but also for other features, where those were genuinely necessary, but missing. Perhaps after the last runner rewrite?

**NOTE:** Asked Fable to code it. Can't say I know "every single line of it", but I'm pretty sure I grasp the "why" and "how" in details.

Caught by a regression in `rs-matter-embassy` [where the new `trouble` introduced ~ 20K+ flash bloat](https://github.com/sysgrok/rs-matter-embassy/pull/67#issuecomment-4980719863). Look at the `nrf52_light_thread` line.


lulf (approved):
Thanks!

PR #669, opened 2026-08-17
Gate LE Connection Update and LE Enable Encryption by controller support

HaoboGu:
and add an extra cfg to gate the central only code

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -101,7 +101,7 @@ default-packet-pool = []
 # Optimization where l2cap SDU reassembly saves some buffer copy.
 l2cap-sdu-reassembly-optimization = []
 
-# Support connection parameter update procedure
+# Enable Connection Parameters Request procedure ([Vol 6] Part B, Section 5.1.7) support.
 connection-params-update = []
 
 # Allow L2CAP SPSU values that are reserved for future use
--- a/host/src/connection.rs
+++ b/host/src/connection.rs
@@ -328,6 +328,10 @@ impl ConnectionParamsRequest {
     /// Accept the connection parameters update request.
     ///
     /// If `params` is `None`, use the parameters requested by the peer.
+    ///
+    /// Applying the parameters takes `HCI_LE_Connection_Update`, and only a Central
+    /// ever receives this request, so this needs the `central` feature.
+    #[cfg(feature = "central")]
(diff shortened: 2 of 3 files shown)


lulf (commented):
I'm not a fan of having multiple APIs for this, but maybe in the cases where you want to make sure it always works , we can expose this.

Perhaps we should name them better to make it clearer which is which and what they do?

HaoboGu:
Yeah I agree. An alternative is to use `connection-params-update` as the feature gate to use LE_Connection_Update for peripherals.

This is a breaking change because for peripherals, the default behavior fallbacks to use L2CAP signaling unless the `update_connection_params` is specified:


```diff

    /// Update connection parameters for this connection.
    pub async fn update_connection_params<T>(
        &self,
        stack: &Stack<'_, T, P>,
        params: &RequestedConnParams,
    ) -> Result<(), BleHostError<T::Error>>
    where
        T: ControllerCmdAsync<LeConnUpdate> + ControllerCmdSync<LeReadLocalSupportedFeatures>,
    {
        let handle = self.handle();
+        #[cfg(any(feature = "central", feature = "connection-params-update"))]
+        {
            // First, check the local supported features to ensure that the connection update is supported.
            let features = stack.host().command(LeReadLocalSupportedFeatures::new()).await?;
            if features.supports_conn_parameters_request_procedure() || self.role() == LeConnRole::Central {
                match stack.host().async_command(into_le_conn_update(handle, params)).await {
                    Ok(_) => return Ok(()),
                    Err(BleHostError::BleHost(crate::Error::Hci(bt_hci::param::Error::UNKNOWN_CONN_IDENTIFIER))) => {
                        return Err(crate::Error::Disconnected.into());
                    }
                    Err(BleHostError::BleHost(crate::Error::Hci(bt_hci::param::Error::UNSUPPORTED_REMOTE_FEATURE))) => {
                        // We tried to send the request as a periperhal but the remote central does not support procedure.
                        // Use the L2CAP signaling method below instead.
                        // This code path should never be reached when acting as a central. If a bugged controller implementation
                        // returns this error code we transmit an invalid L2CAP signal which then is rejected by the remote.
                    }
                    Err(e) => return Err(e),
                }
            }
+        }
```

HaoboGu:
Updated the code

lulf (approved):
Agreed with using feature gating instead 👍

PR #674, opened 2026-08-22
Handle LE Subrate Change event, enable corresponding bits in the LE event mask

HaoboGu:


HaoboGu:
It's ready

tompre:
@HaoboGu You are so fast, man!
I wanted to prepare a pull request as well.

I combined it with the faster connections interval feature flag to make connection rate request work and merged connection params to reflect all interval settings at the same time. Your approach might be cleaner though, as it does not change the api as much.
I added a example for both features, including peripheral requests.

However I based it on the [pull request](https://github.com/embassy-rs/bt-hci/pull/101) for bt-hci, to get type safe host feature flags, but it might be better to merge this now without it.

Feel free to pick the example or the other changes. I'm not free for the next few weeks to work on this:
[tompre/trouble/tree/conn-(sub)-rate-example
](https://github.com/tompre/trouble/tree/conn-(sub)-rate-example
)

HaoboGu:
Yeah, this PR focuses on subrating. I added and removed the connection rate updated part because it's in 6.2 spec. I prefer to add it with a feature gate and in a separate PR. I'm very excited to know that you're working on the new SCI feature, I want it too! Hope to see your PR soon!

host/src/connection_manager.rs
```diff
@@ -1148,6 +1178,7 @@ pub struct ConnectionStorage<P> {
     pub role: LeConnRole,
     pub peer_identity: Identity,
     pub params: ConnParams,
+    pub subrating_params: Option<SubratingParams>,
```
lulf, line 1191: Should these fields also require cfg(feature = subrating) ?

HaoboGu: Yes, nice catch! I'll fix it soon

host/src/host.rs
```diff
@@ -1258,6 +1258,7 @@ impl<'d, C: Controller, P: PacketPool> Runner<'d, C, P> {
             + ControllerCmdSync<SetEventMask>
             + ControllerCmdSync<SetEventMaskPage2>
             + ControllerCmdSync<LeSetEventMask>
+            + ControllerCmdSync<LeSetHostFeature>
```
tompre: We need to feature gate this trait bound, as LeSetHostFeature is optional and might not be implemented by every controller.

tompre (commented):
I prepared a fix similar to the trait bound gating used for the security feature flag.
https://github.com/embassy-rs/trouble/commit/808b09e76905e6f35c592258af7272d02b28b03a

host/src/connection_manager.rs
HaoboGu: Applied subrating feature gate on all related code

host/src/host.rs
HaoboGu: Yes, I added a `SubratingCmds`, just like `SecurityCmds`

HaoboGu:
Added the feature gating. Only feature gate in `ConnectionEvent` is ignored, which is same as `ConnectionRateChanged`(BLE 6.2?)

lulf (approved):

PR #675, opened 2026-08-24
feat(host): set host support flags for ISO and SCI based on crate feature flags

tompre:
This PR adds host feature support flags for Connected Isochronous Streams (ISO, bit 32) and Shorter Connection Intervals (SCI, bit 73 via LeSetHostFeatureV2), gated behind feature flags alongside subrating.

It introduces blanked traits (IsoStreamCmds, ShortConnIntervalCmds) to gate trait bounds depending on the feature flags.

I noticed that the functions Runner::run, Runner::run_with_handler, ControlRunner::run all have a long where clause and could use trouble::Controller instead. Is that on purpose or could it be refactored to make the code more readable?

### DIFF
--- a/host/Cargo.toml
+++ b/host/Cargo.toml
@@ -111,6 +111,9 @@ connection-params-update = []
 # Enable Connection Subrating host support.
 subrating = []
 
+# Enable Shorter Connection Intervals
+shorter-connection-intervals = []
+
 # Allow L2CAP SPSU values that are reserved for future use
 allow-reserved-l2cap-psu = []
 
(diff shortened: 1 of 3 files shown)


lulf (approved):

lulf:
It was intended as a lowest common denominator that libraries based on trouble could use to simplify type signatures.

tompre:
> It was intended as a lowest common denominator that libraries based on trouble could use to simplify type signatures.

Ok, so its for external use interface only and you assure that the trait bounds of the individual functions match using the tester app. That is really useful. :grinning:

PR #679, opened 2026-08-26
feat(examples): add ble_sci_central and ble_sci_peripheral

tompre:
Adds central and peripheral example applications that demonstrate the LE Connection Rate Request procedure (Shorter Connection Intervals, SCI).

The example walks through a four-step negotiation:
1. Central locks the connection interval at exactly 2.5 ms (min=max).
2. Peripheral requests 875 µs (the SCI minimum for nrf52840), which is below the central's minimum of 2.5 ms. The LL rejects the request because the ranges [875 µs, 875 µs] and [2.5 ms, 2.5 ms] do not overlap. Unlike subrate rejections (which always generate a sentinel event), SCI rejection behaviour is controller-defined: the controller may silently drop the request or send ConnectionRateChanged with the previous interval unchanged. The example handles both cases.
3. Central widens the acceptable range to [875 µs, 2.5 ms].
4. Peripheral requests 875 µs again. The intersection is [875 µs, 875 µs] so the LL accepts and both sides receive ConnectionRateChanged.

### DIFF
--- a/examples/apps/Cargo.toml
+++ b/examples/apps/Cargo.toml
@@ -50,3 +50,4 @@ security = [
     "dep:serde"
 ]
 std = []
+shorter-connection-intervals = ["trouble-host/shorter-connection-intervals"]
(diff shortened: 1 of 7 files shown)


lulf:
Same as for #680 - I don't think we need to create an example for this.

tompre:
Fair point, the pull request can be found anyway, so it still might help someone.

PR #680, opened 2026-08-26
feat(examples): add ble_subrate_central and ble_subrate_peripheral

tompre:
Adds central and peripheral example applications that demonstrate the
LE Subrate Request procedure (HCI command 0x007e, Connection::request_conn_subrate).

The example walks through a four-step negotiation:
1. Central sets an subrate factor range of [4, 12].
2. Peripheral requests an out-of-range factor (50). The Link Layer
   rejects the request (ranges do not overlap).
3. Peripheral requests a valid factor (10), which is accepted.
4. Peripheral requests a range [1, 5] that overlaps with the central's range,
   leading to an accepted factor (e.g. 4 or 5).

based on #678

### DIFF
--- a/examples/apps/Cargo.toml
+++ b/examples/apps/Cargo.toml
@@ -50,3 +50,4 @@ security = [
     "dep:serde"
 ]
 std = []
+subrating = ["trouble-host/subrating"]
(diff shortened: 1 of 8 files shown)


lulf:
Although subrating is great, and thanks for your work in adding that, I'm not sure it warrants a specific example for demonstrating it. If we did that with all features, we'd simply get too many examples to maintain.

tompre:
I guess the example can be found via search, so people can use it anyway, but there is no need for maintenance.

PR #681, opened 2026-08-31
fix(gatt): ReadEvent: use correct data length in accept_unprocessed

### LINKED ISSUE #662: `ReadEvent::accept_unprocessed` ignores data.len() and returns a zero length reply
Hello,

In [`ReadEvent::accept_unprocessed`](https://github.com/embassy-rs/trouble/blob/088e09c451177d5db50cf3e58c68d05512265ba0/host/src/gatt.rs#L517-L545ReadEvent) it looks like `data.len()` is ignored when calculating the response length resulting in a zero length reply (the len calculation uses `payload.len()` instead which will be 0 at this point)
```
        let len = payload.len().saturating_sub(offset).min(mtu - 1);
```
This means that `payload.append` generates a 0 byte reply
```
        payload.append(&data.as_gatt()[offset..][..len])?;
```
The len calculation should use `data.len()` instead
```
         let len = data.as_gatt().len().saturating_sub(offset).min(mtu - 1);
```

I found this when trying to use accept_unprocessed.

aswild:
Use the data length when constructing the reply packet.

Fixes: https://github.com/embassy-rs/trouble/issues/662
Fixes: 69a840fe "feat(gatt): Add accept_unprocessed() method to gatt::ReadEvent and gatt::WriteEvent"

### DIFF
--- a/host/src/gatt.rs
+++ b/host/src/gatt.rs
@@ -554,7 +554,7 @@ impl<'stack, P: PacketPool> ReadEvent<'stack, '_, P> {
         // packet-pool buffer and then post-hoc truncated, which can split an
         // entry in half and produce a malformed PDU.
         let mtu = self.data.connection.get_att_mtu() as usize;
-        let len = payload.len().saturating_sub(offset).min(mtu - 1);
+        let len = data.as_gatt().len().saturating_sub(offset).min(mtu - 1);
 
         payload.write(rsp)?;
         payload.append(&data.as_gatt()[offset..][..len])?;


aswild:
Thanks to @paulc for finding this issue and the fix.

I am extremely new to BLE, so hopefully this setup makes sense. I'm using `accept_unprocessed` to implement the Device Information Service, since it seems weird to have to store copies of read-only characteristic data inside the service struct, and it's ugly to have `.try_into().unwrap()` chains inside macro attributes to initialize a `heapless::String` from a string literal.

```rust

const MANUFACTURER_NAME: &str = "...";
const MODEL_NUMBER: &str = "...";
const SERIAL_NUMBER: &str = "...";

struct GattServer {
    device_id_service: DeviceInfoService,
}

#[gatt_service(uuid = service::DEVICE_INFORMATION)]
struct DeviceInfoService {
    #[characteristic(uuid = characteristic::MANUFACTURER_NAME_STRING, read)]
    manufacturer_name: (),
    #[characteristic(uuid = characteristic::MODEL_NUMBER_STRING, read)]
    model_number: (),
    #[characteristic(uuid = characteristic::SERIAL_NUMBER_STRING, read)]
    serial_number: (),
    #[characteristic(uuid = characteristic::SOFTWARE_REVISION_STRING, read)]
    software_revision: (),
}

impl GattServer<'_> {
    fn handle_gatt_event<'stack, 'server, P: PacketPool>(
        &self,
        event: GattEvent<'stack, 'server, P>,
    ) -> Result<Reply<'stack, P>, trouble_host::Error> {
        match event {
            GattEvent::Read(event) => {
                // what are we reading?
                let handle = event.handle();
                let svc = &self.device_id_service;
                if handle == svc.manufacturer_name.handle {
                    info!("GATT read manufacturer name");
                    event.accept_unprocessed(MANUFACTURER_NAME)
                } else if handle == svc.model_number.handle {
                    info!("GATT read model number");
                    event.accept_unprocessed(MODEL_NUMBER)
                } else if handle == svc.serial_number.handle {
                    info!("GATT read serial number");
                    event.accept_unprocessed(SERIAL_NUMBER)
                } else if handle == svc.software_revision.handle {
                    info!("GATT read software revision");
                    event.accept_unprocessed(env!("CARGO_PKG_VERSION"))
                } else {
                    event.reject(AttErrorCode::ATTRIBUTE_NOT_FOUND)
                }
            }
            _ => event.accept(),
        }
    }
}
```

lulf:
Thanks!

PR #685, opened 2026-09-22
fix(host): fall back to L2CAP for peripheral parameter updates

cedrivard:
## Summary

Some controllers advertise support for the LE Connection Parameters Request Procedure but reject the HCI LE Connection Update command. CYW43439 exhibits this behavior by returning Invalid LMP LL Parameters.

This change:

- falls back to L2CAP for peripheral connections when HCI returns Unsupported Remote Feature or Invalid LMP LL Parameters
- preserves HCI errors for central connections
- validates requested connection parameters before selecting either update procedure
- clarifies why the fallback must remain peripheral-only

## Hardware verification

Verified on a CYW43439-based RedBoard with Linux/BlueZ as the central. The peripheral requested a 30 ms connection interval, latency 0, and a 4 second supervision timeout. The resulting connection reported those exact parameters, completed the phase challenge, and produced no fatal signatures.

## Verification

- `cargo +1.97 fmt --check`
- `cargo +1.97 test --locked --lib` (24 passed)
- `cargo +1.97 test --locked --lib --no-default-features --features peripheral,connection-params-update,default-packet-pool` (11 passed)
- release build with `peripheral,connection-params-update,default-packet-pool`
- release build with `central,default-packet-pool`


lulf:
You can also workaround this by disabling the `connection-params-update` feature alltogether, which will cause it to use l2cap always. This seems like a bug in the cyw43 controller though, so it feels like just disabling `connection-params-feature` when using this controller is the appropriate fix.

cedrivard:
Thanks, agreed, disabling `connection-params-update` is the cleaner fix here. I tried it on the board, and the runtime parameter update still works over L2CAP. I’ll withdraw the PR.

